PR #4, opened 2018-09-20
main: add a way for the app to persist

cosimoc:
Make development easier.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -20,8 +20,12 @@ pkg.require({
     HackToolbox: '0',
 });
 
+const {GLib} = imports.gi;
 const {HackToolboxApplication} = imports.app;
 
 function main(argv) {
-    return (new HackToolboxApplication()).run(argv);
+    let app = new HackToolboxApplication();
+    if (GLib.getenv('HACK_TOOLBOX_PERSIST'))
+        app.hold();
+    return app.run(argv);
 }


src/main.js
```diff
@@ -20,8 +20,12 @@ pkg.require({
     HackToolbox: '0',
 });
 
+const GLib = imports.gi.GLib;
```
ptomato: The new hotness is to use destructuring: `const {GLib} = imports.gi;`

src/main.js
```diff
@@ -20,8 +20,12 @@ pkg.require({
...
 const {HackToolboxApplication} = imports.app;
 
 function main(argv) {
-    return (new HackToolboxApplication()).run(argv);
+    let app = new HackToolboxApplication();
+    if (GLib.getenv('HACK_TOOLBOX_PERSIST'))
+        app.hold();
```
ptomato, line 29: Seems fine as a quick hack even though we never release it. It's unfortunate the app class is hardcoded to have the IS_SERVICE flag set, but fixing that would be more trouble than this small PR warrants.

ptomato:
All optional comments, feel free to merge.

src/main.js
cosimoc: Yeah, that was too much yak shaving for something this simple. We use this pattern in other similar apps like the discovery feed.

src/main.js
cosimoc: Fixed.

PR #5, opened 2018-09-21
Framework app control panel

ptomato:
This is the whole of the control panel except for some sound packs which I haven't curated yet. (We still need two click sound packs and three hover sound packs, of 6 sounds each.)

The code view is not implemented yet.

https://phabricator.endlessm.com/T23625


src/framework/appWindow.js
```diff
@@ -0,0 +1,24 @@
...
+const {GObject} = imports.gi;
+
+const {RaControlPanel} = imports.framework.controlPanel;
+const {RaModel} = imports.framework.model;
+const {ToolboxWindowBase} = imports.window;
+
+var RaAppWindow = GObject.registerClass(class RaAppWindow extends ToolboxWindowBase {
+    _init(props = {}) {
+        super._init(props);
+        this.maximize();
```
cosimoc: I wouldn't do this here... The flip to hack code in the shell should already take care of synchronizing the toolbox window geometry with the original one.

src/framework/model.js
```diff
@@ -0,0 +1,211 @@
...
+
+    set hover_sound(value) {
+        if ('_hoverSound' in this && this._hoverSound === value)
+            return;
+        this._hoverSound = value;
+        this.notify('hover-sound');
+
+        if (value !== 'none') {
+            this._clickSound = 'none';
+            this.notify('click-sound');
```
cosimoc, line 211: Why can't a click sound and a hover sound be both set at the same time?

src/meson.build
```diff
@@ -55,3 +55,5 @@ gnome.generate_gir(main_library, extra_args: ['--warn-all', '--warn-error'],
     sources: introspection_sources, symbol_prefix: 'hack_toolbox')
 
 install_headers(installed_headers, subdir: api_name)
+
+subdir('framework')
```
cosimoc, line 61: This should probably go in the previous commit?

src/framework/textFilters.js
```diff
@@ -0,0 +1,255 @@
...
+        const middle = Utils.shuffle(letters.slice(1, -1), sequence);
+        const end = letters[letters.length - 1];
+        return beginning + middle.join('') + end;
+    });
+    return words.join(' ');
+}
+
+function zalgo(text) {
+    // Thanks to Michael S. Kaplan:
+    // http://blogs.msdn.com/b/michkap/archive/2006/02/17/533929.aspx
```
cosimoc: Page does not seem to exist :(

src/framework/textFilters.js
```diff
@@ -0,0 +1,255 @@
...
+    // 100. randomization 70%, height goes from 30 to 100.
+    const RANDOMIZATION = 100;
+
+    let newText = '';
+    [...text].forEach(letter => {
+        let newChar = letter;
+
+        // Middle
+        // Put just one of the middle characters there, or it gets crowded
+        const middleLen = ZALGO_DIACRITICS.middle.length;
```
cosimoc, line 228: Should this not be length -1?

src/framework/model.js
```diff
@@ -208,4 +227,86 @@ var RaModel = GObject.registerClass({
...
+        const manifest = Gen.generateGResource(this);
+
+        const tmpDir = Gio.File.new_for_path(GLib.get_user_runtime_dir());
+        const logoResource = Gio.File.new_for_uri(`resource://${this._logoGraphic}`);
+        const logo = tmpDir.get_child('logo');
+        const manifestXml = tmpDir.get_child('hacking.gresource.xml');
+        const gresource = tmpDir.get_child('hacking.gresource');
+
+        const logoInputStream = await logoResource.read_async(_PRIO, null);
+        const logoOutputStream = await logo.replace_async(null, false,
```
cosimoc: Couldn't you simply use `Gio.File.copy()` for this?

src/framework/model.js
```diff
@@ -208,4 +227,86 @@ var RaModel = GObject.registerClass({
...
+            null, false, Gio.FileCreateFlags.NONE, null);
+
+        const proc = new Gio.Subprocess({
+            argv: ['glib-compile-resources', '--target', gresource.get_path(),
+                '--sourcedir', tmpDir.get_path(), manifestXml.get_path()],
+        });
+        proc.init(null);
+        await proc.wait_check_async(null);
+
+        // Ah, geez, this is terrible
```
cosimoc: What is this doing exactly?

src/framework/banner/hackableSearch.js
```diff
@@ -0,0 +1,52 @@
+// Copyright 2015 Endless Mobile, Inc.
```
cosimoc: 2018

cosimoc (changes requested):
Looks pretty good! This is a first pass review, as there's quite a bit of code. @tchx84 or @xjuan can perhaps help with the next round...

src/framework/model.js
ptomato: No fundamental reason, but this made it simpler to implement and I also thought it would make it difficult to play a melody on the piano, for example, if you were also getting hover sounds in between. I can remove that restriction, but if so we should create another issue for it.

src/meson.build
ptomato: There was no meson.build file needed in the framework directory until this commit — it's only needed to build the GResource file with the extra framework modules.

src/framework/textFilters.js
ptomato: I see ... I looked it up on archive.org and it's really only tangentially related, so I'll just delete this.

src/framework/textFilters.js
ptomato: No, I don't think so - `Math.floor(Math.random() * middleLen)` will return an index from 0 to `middleLen - 1`, so it works out okay.

src/framework/model.js
ptomato: So, this is the only way I could find to open a file and get an FD using GJS ... it uses Gio.Subprocess to run the command `cat myfile`, and grabs the FD from its stdout pipe.

src/framework/model.js
cosimoc: Sounds good

ptomato:
OK, finally made the changes based on the above comments. I'll rebase the `controlpanel-cosimoc` branch on top of this.

ptomato:
Added the internal C library for opening files.

PR #6, opened 2018-09-26
Controlpanel cosimoc

cosimoc:
Some fixes for the controlpanel branch that I had to make for the flip-to-hack integration.


ptomato:
CR pass, these can be merged into the other pull request

PR #7, opened 2018-09-27
window: move flip-back action to window

cosimoc:
Since HackToolbox can have multiple.

### DIFF
--- a/src/window.js
+++ b/src/window.js
@@ -24,7 +24,26 @@ var ToolboxWindowBase = GObject.registerClass({
         this.hack_toolbox_skeleton = this._createHackToolboxSkeletonOnPath(objectPath);
         this.hackable_proxy = this._maybeGetHackableProxySync();
 
-        this.application.enableFlipBack = false;
+        this._flipBack = new Gio.SimpleAction({
+            name: 'flip-back',
+        });
+        this._flipBack.connect('activate', this._onFlipBack.bind(this));
+    }
+
+    _onFlipBack(action, parameterVariant) {
+        log(`Call flip-back for ${this.target_bus_name}, ${this.target_object_path}`);
+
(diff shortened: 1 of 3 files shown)


cosimoc:
@ptomato We discussed why this was necessary yesterday.

src/window.js
```diff
@@ -24,7 +24,34 @@ var ToolboxWindowBase = GObject.registerClass({
...
-        this.application.enableFlipBack = false;
+        this._flipBack = new Gio.SimpleAction({
+            name: 'flip-back',
+            parameter_type: new GLib.VariantType('(ss)'),
+        });
+        this._flipBack.connect('activate', this._onFlipBack.bind(this));
+    }
+
+    _onFlipBack(action, parameterVariant) {
+        this.application.hold();
```
ptomato: I think we don't need to hold/release the application now, since the window will keep the application alive.

src/window.js
```diff
@@ -24,7 +24,34 @@ var ToolboxWindowBase = GObject.registerClass({
         this.hack_toolbox_skeleton = this._createHackToolboxSkeletonOnPath(objectPath);
         this.hackable_proxy = this._maybeGetHackableProxySync();
 
-        this.application.enableFlipBack = false;
+        this._flipBack = new Gio.SimpleAction({
+            name: 'flip-back',
+            parameter_type: new GLib.VariantType('(ss)'),
```
ptomato: I'm not sure we need the parameters anymore, since passing the parameters to the app was necessary to decide what window to route the action too. Now we talk directly to the window already.

ptomato (changes requested):
Looks good, I think a few more simplifications are possible though.

ptomato (approved):

PR #8, opened 2018-09-27
Refactor toolbox to be a widget

ptomato:
Previously the toolboxes were separate windows, but we want to have them
be widgets inside one common window instead, so that they can have
common styling without reimplementing the same thing for each toolbox.

https://phabricator.endlessm.com/T23902


tchx84 (approved):

tchx84:
LGTM

PR #11, opened 2018-10-02
More codeview

ptomato:
Some more changes to the codeview from today.

https://phabricator.endlessm.com/T23902


src/codeview.js
```diff
@@ -124,6 +141,20 @@ var Codeview = GObject.registerClass({
         this._helpMessage.pointingTo = area;
         this._helpMessage.relativeTo = this._view;
         this._helpMessage.get_style_context().add_class('error');
+
+        if (this._fixHandler)
+            this._fixButton.disconnect(this._fixHandler);
+        this._fixButton.hide();
+        print('deciding whether to show fix button');
```
tchx84, line 148: debug leftover?

tchx84 (approved):
I did not test the integration with the shell, but the toolbox itself and codeview (plus error handling, etc) works great!

src/codeview.js
tchx84: Well, I guess we merge it as-is and we remove those later.

PR #12, opened 2018-10-02
T23807

tchx84:
https://phabricator.endlessm.com/T23807

### DIFF
--- a/src/framework/buttonGroup.js
+++ b/src/framework/buttonGroup.js
@@ -1,9 +1,11 @@
 /* exported ButtonGroup */
 
-const {GObject} = imports.gi;
+const {GObject, GLib} = imports.gi;
 
 var ButtonGroup = GObject.registerClass({
     Properties: {
+        index: GObject.ParamSpec.uint('index', 'index', '',
+            GObject.ParamFlags.READWRITE, 0, GLib.MAXUINT32, 0),
         value: GObject.ParamSpec.string('value', 'Value', '',
             GObject.ParamFlags.READWRITE, ''),
     },
@@ -19,6 +21,22 @@ var ButtonGroup = GObject.registerClass({
         this._updateValue();
     }
 
+    get index() {
+        return this._index;
+    }
+
+    set index(index) {
+        if ('_index' in this && this._index === index)
+            return;
+        if (index >= this._enumValues.length)
+            throw new Error(`Invalid index ${index}`);
+        this._index = index;
+        this._value = this._enumValues[this._index];
+        this._updateUI();
+        this.notify('index');
+        this.notify('value');
+    }
+
     get value() {
         return this._value;
     }
@@ -29,8 +47,10 @@ var ButtonGroup = GObject.registerClass({
         if (!this._enumValues.includes(value))
             throw new Error(`Invalid value ${value}; expected ${this._enumValues}`);
         this._value = value;
+        this._index = this._enumValues.indexOf(value);
         this._updateUI();
         this.notify('value');
+        this.notify('index');
     }
 
     _updateValue() {
--- a/src/hackyballs/controlpanel.js
+++ b/src/hackyballs/controlpanel.js
@@ -1,0 +1,55 @@
+/* exported HBControlPanel */
+
+const {GObject, Gtk} = imports.gi;
+const {HBLevel1} = imports.hackyballs.level1;
+const {HBLevel2} = imports.hackyballs.level2;
+const {HBLevel3} = imports.hackyballs.level3;
+const {Lockscreen} = imports.lockscreen;
+
+var HBControlPanel = GObject.registerClass({
+    GTypeName: 'HBControlPanel',
+    Template: 'resource:///com/endlessm/HackToolbox/hackyballs/controlpanel.ui',
+    InternalChildren: [
+        'panelLevel1',
+        'panelLevel2',
+        'panelLevel3',
+    ],
+}, class HBControlPanel extends Gtk.Box {
+    _init(props = {}) {
+        super._init(props);
+
+        this._level1 = new HBLevel1({visible: true});
+        this._level2 = new HBLevel2({visible: true});
+        this._level3 = new HBLevel3({visible: true});
+        this._level2lock = new Lockscreen({visible: true});
+        this._level3lock = new Lockscreen({visible: false});
+
+        this._level2lock.add(this._level2);
+        this._level3lock.add(this._level3);
+        this._panelLevel1.add(this._level1);
+        this._panelLevel2.add(this._level2lock);
+        this._panelLevel3.add(this._level3lock);
+    }
+
+    bindModel(model) {
+        this._level1.bindModel(model);
+        this._level2.bindModel(model);
+        this._level3.bindModel(model);
+        this._model = model;
+    }
+
+    _onUnlockStateChanged(win) {
+        const [, l2, l3] = win.getUnlockState();
+        if (l2) {
+            this._level2lock.locked = false;
+            this._level3lock.show_all();
+        }
+        if (l3)
+            this._level3lock.locked = false;
+    }
+
+    bindWindow(win) {
+        win.connect('unlock-state-changed', this._onUnlockStateChanged.bind(this));
+        this._onUnlockStateChanged(win);
+    }
+});
--- a/src/hackyballs/level1.js
+++ b/src/hackyballs/level1.js
@@ -1,0 +1,32 @@
+/* exported HBLevel1 */
+
+const {GObject, Gtk} = imports.gi;
+const {ButtonGroup} = imports.framework.buttonGroup;
+
+var HBLevel1 = GObject.registerClass({
+    GTypeName: 'HBLevel1',
+    Template: 'resource:///com/endlessm/HackToolbox/hackyballs/level1.ui',
+    InternalChildren: [
+        'background0',
+        'background1',
+        'background2',
+        'background3',
+        'background4',
+    ],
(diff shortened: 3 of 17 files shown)


src/hackyballs/controlpanel.js
```diff
@@ -0,0 +1,44 @@
...
+const {GObject, Gtk} = imports.gi;
+const {HBLevel1} = imports.hackyballs.level1;
+const {HBLevel2} = imports.hackyballs.level2;
+const {HBLevel3} = imports.hackyballs.level3;
+
+var HBControlPanel = GObject.registerClass({
+    GTypeName: 'HBControlPanel',
+    Template: 'resource:///com/endlessm/HackToolbox/hackyballs/controlpanel.ui',
+    InternalChildren: [
+        'buttonReset',
```
ptomato: I think it would be better to make the reset button a permanent part of the toolbox window, and have it emit a signal which is handled by your control panel.

src/hackyballs/controlpanel.js
```diff
@@ -0,0 +1,44 @@
...
+        this._panelLevel3.add(this._level3);
+        this._buttonReset.connect('clicked', () => {
+            this._onReset();
+        });
+    }
+
+    _onReset() {
+        const props = GObject.Object.list_properties.call(this._model.constructor.$gtype);
+        props.forEach(prop => {
+            this._model.set_property(prop.get_name(), prop.default_value);
```
ptomato: `this._model[prop.get_name()] = prop.default_value`?

src/hackyballs/level1.js
```diff
@@ -0,0 +1,32 @@
...
+        'background2',
+        'background3',
+        'background4',
+        'background5',
+    ],
+}, class HBLevel1 extends Gtk.Grid {
+    _init(props = {}) {
+        super._init(props);
+        this._backgroundGroup = new ButtonGroup({
+            0: this._background1,
```
ptomato: Since you are using `index` anyway to bind, you can give these any name you want. It might be better to be more descriptive.

src/hackyballs/level3.js
```diff
@@ -0,0 +1,119 @@
...
+        this._getModelProps().forEach(pspec => {
+            const propName = pspec.get_name();
+            const propType = typeof pspec.default_value;
+            if (!scope[propName])
+                return;
+            if (scope[propName] !== null && typeof scope[propName] !== propType) {
+                errors.push(this._errorRecordAtAssignmentLocation(
+                    propName,
+                    `Unknown value ${scope[propName]}: value must be a ${propType}`,
+                    pspec.default_value
```
ptomato, line 95: Is `pspec.default_value` the right value in all cases? In my case, a `Gdk.RGBA` (i.e. GBoxed) property's default value is always `null`, so I had to provide my own defaults...

src/hackyballs/model.js
```diff
@@ -0,0 +1,170 @@
...
+            const proxyProp = map[localProp] ? map[localProp] : localProp;
+            this._bindLocalToProxy(localProp, proxyProp);
+            invertMap[proxyProp] = localProp;
+        });
+
+        this._proxy.connectSignal('ObjectNotify', (p, s, [o, proxyProp, ret]) => {
+            const value = ret.deep_unpack();
+            const localProp = invertMap[proxyProp] ? invertMap[proxyProp] : proxyProp;
+
+            /* XXX use GObject.param_values_cmp */
```
ptomato: Is this meant to be addressed before merging?

ptomato (changes requested):
I'm having trouble running it, I will comment on the Phabricator issue.

src/hackyballs/controlpanel.js
tchx84: what do you mean by toolbox window? You mean the fake one? Because the real one is not connected to the fake one.

src/hackyballs/controlpanel.js
tchx84: Remember the issue I mentioned in SF? Every time I use `this._model[name] = value`, the `notify` signal never gets emitted and everything gets out of sync ;/

src/hackyballs/level3.js
tchx84: For hackyballs, yes. But is a valid point for whenever we try to generalize this code.

src/hackyballs/model.js
tchx84: It was my intention, but there is still one issue with how that function behaves, e.g. it will return '0' when the type passed is not strictly the same, like '10' and '10.1' it will still return '0' (and will print a CRITICAL). In other words, it only behaves properly, when the two values are strictly the same type.

So, I tried to complement it with `param_value_validate` but something weird happening with it as well.

```
const {GObject, Gtk} = imports.gi;

Gtk.init(null);

const w = new Gtk.Box();
const pspec = GObject.Object.find_property.call(w.constructor.$gtype, 'visible');
print(pspec);
print(pspec.owner_type);
print(pspec.default_value);
print(GObject.param_value_validate(pspec, pspec.default_value));
print(GObject.param_values_cmp(pspec, pspec.default_value, true));
print(GObject.param_values_cmp(pspec, pspec.default_value, false));
print(GObject.param_values_cmp(pspec, pspec.default_value, 'true'));
```

http://devdocs.baznga.org/gobject20~2.44p/gobject.param_values_cmp
http://devdocs.baznga.org/gobject20~2.44p/gobject.param_value_validate

So I would prefer not get blocked on this right now.

ptomato: I think I figured out that `param_values_cmp` and `param_value_validate` are currently unusable from GJS, because you don't have control over what type of GValue gets created, but to use the function correctly, the passed-in GValues must adhere to certain types. And there's no way to express that in g-i annotations.

I think it might be possible to make this function work correctly by overriding `GObject.ParamSpec.prototype.vfunc_values_cmp()` and just reimplementing the logic of the C function in JS, but I agree to not block on that. Would you mind opening a bug for GJS?

tchx84: Yes, ill report it.

tchx84: https://gitlab.gnome.org/GNOME/gjs/issues/210

src/framework/buttonGroup.js
```diff
@@ -1,9 +1,11 @@
 /* exported ButtonGroup */
 
-const {GObject} = imports.gi;
+const {GObject, GLib} = imports.gi;
 
 var ButtonGroup = GObject.registerClass({
     Properties: {
+        index: GObject.ParamSpec.uint('index', 'index', '',
```
ptomato, line 7: I would split this change out into a separate commit.

src/hackyballs/model.js
```diff
@@ -0,0 +1,177 @@
+/* exported HBModelBase, HBModelGlobal, HBModelObject */
+
+const {Gio, GLib, GObject} = imports.gi;
+
+const ViewName = 'view.JSContext.globalParameters';
+const BusName = 'com.endlessm.hackyballs';
+const AppObjectPath = '/com/endlessm/hackyballs';
+const KnowledgeControlIface = `
```
ptomato: Call this `ClippyIface` or something similar.

src/hackyballs/model.js
ptomato: I would remove the `XXX` then, to show that it's not something we really need to solve

src/hackyballs/model.js
```diff
@@ -0,0 +1,177 @@
...
+      <arg type='s' name='property' />
+      <arg type='v' name='value' />
+    </signal>
+  </interface>
+</node>
+`;
+
+const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
+
+var HBModelBase = GObject.registerClass({
```
ptomato, line 37: Why is it needed to split it into a base model and a derived model?

ptomato (changes requested):
Just a few comments.

I pushed a small change to my branch as well, so it would be good to rebase on T23957 as well.

src/hackyballs/model.js
tchx84: I will need other models for different hack able things, and I need to be able to pass mapping data between app variables and my local variables.

src/hackyballs/sliders.js
```diff
@@ -0,0 +1,32 @@
...
+        'adjustmentRadius',
+        'adjustmentGravity',
+        'adjustmentCollision',
+        'adjustmentFriction',
+        'adjustmentSocial0',
+        'adjustmentSocial1',
+        'adjustmentSocial2',
+    ],
+}, class HBSliders extends Gtk.Grid {
+    _init(props = {}) {
```
ptomato: optional: The `_init` function can be left out if it only chains up.

src/hackyballs/controlpanel.js
```diff
@@ -0,0 +1,52 @@
...
+
+    bindModel(model) {
+        this._level1.bindModel(model);
+        this._level2.bindModel(model);
+        this._level3.bindModel(model);
+        this._model = model;
+    }
+
+    bindWindow(win) {
+        win.connect('unlock-state-changed', () => {
```
ptomato: I made the same mistake in my original lockscreen PR and had to fix it later; when you bind the window you have to sync the initial lockscreens states with the window's unlock state, or else you will get a wrong state when you restart the app without restarting the hack toolbox service.

See commit 1b1c91ab7bc2ccded01b07cc4f66c0e838ac0d28

ptomato (changes requested):
Only one important comment

ptomato (approved):

PR #13, opened 2018-10-03
Lock screens

ptomato:
https://phabricator.endlessm.com/T23929

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -1,0 +1,78 @@
+/* exported Lockscreen */
+
+const {GObject, Gtk} = imports.gi;
+
+const _css = `
+frame {
+    background: -gtk-icontheme('dialog-password'),
+                linear-gradient(to right, rgba(61, 70, 81, 1) 0%,
+                                          rgba(61, 70, 81, 1) 50%,
+                                          rgba(61, 70, 81, 0) 51%,
+                                          rgba(61, 70, 81, 0) 100%),
+                linear-gradient(to right, rgba(61, 70, 81, 0) 0%,
+                                          rgba(61, 70, 81, 0) 50%,
+                                          rgba(61, 70, 81, 1) 51%,
+                                          rgba(61, 70, 81, 1) 100%);
+    background-repeat: no-repeat;
+    background-position: center, -600px, 600px;
+    opacity: 0;
+    transition-property: background-position, opacity;
+    transition-duration: 0.5s, 0.75s;
+    transition-timing-function: cubic-bezier(.5, 0, 1, .5), linear;
+}
+
+frame.locked {
+    background-position: center, 0px, 0px;
+    opacity: 1;
+}
+`;
+
+var Lockscreen = GObject.registerClass({
+    Properties: {
+        locked: GObject.ParamSpec.boolean('locked', 'Locked', '',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT,
+            true),
+    },
+}, class Lockscreen extends Gtk.Overlay {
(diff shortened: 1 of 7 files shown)


src/lockscreen.js
```diff
@@ -0,0 +1,81 @@
+/* exported Lockscreen */
+
+const {GObject, Gtk} = imports.gi;
+
+const _css = `
+.lockscreen {
+    background: url('/usr/share/icons/Adwaita/256x256/status/dialog-password.png'),
```
cosimoc: I know this is a stub, but would `-gtk-icontheme` work instead of hardcoding a path?

src/lockscreen.js
```diff
@@ -0,0 +1,81 @@
...
+        this._updateUI();
+        this.notify('locked');
+    }
+
+    _updateUI() {
+        if (!this._overlay)
+            return;
+
+        const style = this._overlay.get_style_context();
+        style.add_class('lockscreen');
```
cosimoc: This should not be required, the style class is already always added in `_init()`

src/lockscreen.js
```diff
@@ -0,0 +1,81 @@
...
+        super._init(props);
+
+        this._overlay = new Gtk.Frame({
+            expand: true,
+            visible: true,
+        });
+        this.add_overlay(this._overlay);
+
+        const style = this._overlay.get_style_context();
+        style.add_class('lockscreen');
```
cosimoc: Instead of a style class, you should be able to set a CSS name for this class; gjs has a way to do that directly as part of the class definition.

src/window.js
```diff
@@ -64,6 +72,23 @@ var ToolboxWindow = GObject.registerClass({
...
+
+    unlock() {
+        if (!(this.target_bus_name in unlockState))
+            unlockState[this.target_bus_name] = [true];
+        const index = unlockState[this.target_bus_name].findIndex(state => !state);
+        unlockState[this.target_bus_name][index] = true;
+        this.emit('unlock-state-changed');
+
+        // Update state of the common lockscreen in this window
+        this._lockscreen.locked = !this.getUnlockState()[0];
```
cosimoc, line 105: This belongs to a different commit

src/window.js
```diff
@@ -70,6 +70,12 @@ var ToolboxWindow = GObject.registerClass({
         this._lockscreen.add(this._toolbox_frame);
 
         this.get_style_context().add_class('toolbox-surrounding-window');
+
+        const accelGroup = new Gtk.AccelGroup();
+        accelGroup.connect('r'.charCodeAt(),
+            Gdk.ModifierType.SHIFT_MASK | Gdk.ModifierType.CONTROL_MASK, 0,
+            this._onKeyCombo.bind(this));
+        this.add_accel_group(accelGroup);
```
cosimoc: It's probably better to make this a GAction in the window action group, in case we ever need to unlock this remotely (and it's more natural to add a keyboard shortcut for it)

cosimoc (changes requested):

src/lockscreen.js
ptomato: I don't think so, as we don't call `gtk_render_icon()` anywhere in this widget.

src/lockscreen.js
ptomato: I want the style class on the overlay widget, instead of the class itself. However, since the style provider only applies to the one widget anyway, I'll just use its existing CSS name (`frame`).

src/window.js
ptomato: Fixed.

ptomato:
Should be all fixed now.

PR #16, opened 2018-10-04
framework: Dim background of toolbox

ptomato:
To indicate that the framework apps do not live-update, we dim the
background of the toolbox window.

https://phabricator.endlessm.com/T23976

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -33,3 +33,7 @@ popover.error {
     background-color: #ffb1aa;
     padding: 12px;
 }
+
+.framework.toolbox-surrounding-window {
+    background-color: rgba(0, 0, 0, 0.6);
+}
--- a/src/window.js
+++ b/src/window.js
@@ -17,6 +17,15 @@ function _shouldEnableFlipBack(targetBusName) {
     }
 }
 
+function _extraToplevelCssClass(targetBusName) {
+    switch (targetBusName) {
+    case 'com.endlessm.dinosaurs.en':
+        return 'framework';
+    default:
+        return null;
+    }
+}
+
 var ToolboxWindow = GObject.registerClass({
     Properties: {
         'target-bus-name': GObject.ParamSpec.string('target-bus-name',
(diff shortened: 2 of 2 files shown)


src/framework/toolbox.js
```diff
@@ -16,6 +16,11 @@ var FrameworkToolbox = GObject.registerClass(class FrameworkToolbox extends Gtk.
         this._model = new RaModel();
 
         this._controlPanel.bindModel(this._model);
+
+        const provider = new Gtk.CssProvider();
+        provider.load_from_resource('/com/endlessm/HackToolbox/framework/application.css');
+        Gtk.StyleContext.add_provider_for_screen(Gdk.Screen.get_default(),
+            provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION + 1);
```
cosimoc: Why is this added for the whole screen? It should only be tied to the lifecycle of the framework toolbox, so should either be added to its style context, or removed when the toolbox is destroyed.

cosimoc (changes requested):
One comment below; I think this is OK for now to be handled here, but in the future we may want to consider moving some of this shading to the shell.

src/framework/toolbox.js
ptomato: I was assuming that when the theming for the toolboxes is done, we will want to put more CSS rules in here that only apply to the framework toolbox.
I guess the best thing to do is to remove it when the window is destroyed.

cosimoc: Yeah, that would probably work. The toolbox process is reused for different windows that are not just for the framework, that's why I wanted it to only apply to the framework toolbox.

ptomato: I was thinking that last night after I wrote this comment.
So my suggestion wouldn't work. We should add a `.framework` CSS class to the toplevel window and add the CSS rules to the main `application.css` instead.

ptomato:
I took a different approach now.

PR #18, opened 2018-10-06
Reorganize framework hacking panel

ptomato:
https://phabricator.endlessm.com/T23977

### DIFF
--- a/src/framework/customModules/banner/hackableDynamic.js
+++ b/src/framework/customModules/banner/hackableDynamic.js
@@ -10,6 +10,7 @@ const Module = imports.framework.interfaces.module;
 const DynamicLogo = imports.framework.widgets.dynamicLogo;
 const FormattableLabel = imports.framework.widgets.formattableLabel;
 const TextFilters = custom_modules.textFilters;
+const TextProcessing = custom_modules.textProcessing;
 const Utils = imports.framework.utils;
 
 void DynamicLogo;
@@ -36,6 +37,14 @@ var HackableDynamic = new Module.Class({
         textfilter: GObject.ParamSpec.string('textfilter', 'Text filter', '',
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             'normal'),
+        rotation: GObject.ParamSpec.uint('rotation', 'Rotation',
+            'Number of positions to advance each letter in the string',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            0, 25, 0),
+        decodefunc: GObject.ParamSpec.string('decodefunc', 'Decode Function',
+            'Source code of function to evaluate to decode the text',
+            GObject.ParamFlags.WRITABLE | GObject.ParamFlags.CONSTRUCT,
+            ''),
     },
 
     Template: 'resource:///com/endlessm/HackToolbox/CustomModules/banner/hackableDynamic.ui',
@@ -86,8 +95,30 @@ var HackableDynamic = new Module.Class({
         return this._subtitle || '';
     },
 
+    get rotation() {
+        return this._rotation;
+    },
+
+    set rotation(value) {
+        this._rotation = value;
+    },
+
+    get decodefunc() {
+        throw new Error('property not readable');
+    },
+
+    set decodefunc(value) {
+        if (!value) {
+            this._decodefunc = null;
+            return;
+        }
+        // eslint-disable-next-line no-new-func
+        this._decodefunc = new Function('letter', value);
+    },
+
     _processText(text) {
-        const decomposedText = text.normalize('NFKD');
-        return TextFilters[this.textfilter](decomposedText);
+        const caesar = TextProcessing.caesar(text, this._rotation,
+            this._decodefunc);
+        return TextFilters[this.textfilter](caesar);
     },
 });
--- a/src/framework/level1.js
+++ b/src/framework/level1.js
@@ -1,17 +1,33 @@
 /* exported FrameworkLevel1 */
 
 const {GObject, Gtk} = imports.gi;
+const Gettext = imports.gettext;
+
 const {logoIDToResource, LogoImage, VALID_LOGOS} = imports.framework.logoImage;
 const {PopupMenu} = imports.popupMenu;
+
+const _ = Gettext.gettext;
 
 var FrameworkLevel1 = GObject.registerClass({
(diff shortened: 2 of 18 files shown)


src/framework/level1.js
```diff
@@ -34,6 +28,7 @@ var FrameworkLevel1 = GObject.registerClass({
         model.accent_color = this._accentColorButton.rgba;
         model.info_color = this._infoColorButton.rgba;
         model.logo_color = this._logoColorButton.rgba;
+        model.logo_graphic = 'dinosaur';
```
cosimoc: Should this not use `this._logoGroup.value`?

src/framework/level1.js
```diff
@@ -2,30 +2,24 @@
...
     Template: 'resource:///com/endlessm/HackToolbox/framework/level1.ui',
     InternalChildren: ['accentColorButton', 'fontButton', 'fontSizeAdjustment',
-        'infoColorButton', 'logoButton', 'logoChoices', 'logoColorButton',
-        'logoMenu', 'mainColorButton'],
+        'infoColorButton', 'logoButton', 'logoColorButton', 'mainColorButton'],
 }, class FrameworkLevel1 extends Gtk.Grid {
     _init(props = {}) {
         super._init(props);
 
+        const logoChoices = {};
```
cosimoc, line 32: I wonder if it wouldn't be easier to instead have this as a `Map`. That way, if you stored it in the menu instead of separately storing the keys and values, you could potentially avoid the double scan in `_updateForPropValue()`

cosimoc (changes requested):

src/framework/level1.js
ptomato: Good point.

src/framework/level1.js
ptomato: Fixed in `popupMenu`.

ptomato:
The "hyperlinks" and "cipher" controls are still coming, but here's the existing commits revised.

src/framework/level1.js
cosimoc: Still not fixed though?

src/popupMenu.js
```diff
@@ -0,0 +1,92 @@
...
+
+    _onChoicesActivated(widget, child) {
+        const choice = child.get_child();  // child of the FlowBoxChild
+        this._updateForPropValue(choice[this._itemProp]);
+        this._menu.popdown();
+    }
+
+    _updateForPropValue(propValue) {
+        this._selected[this._itemProp] = propValue;
+        this.value = this._propMap.get(propValue);
```
cosimoc: This line will also cause the one above to be executed as part of the call to `_updateForEnumValue()` through the setter, so the line above could be removed.

src/popupMenu.js
```diff
@@ -0,0 +1,92 @@
...
+const {GObject, Gtk} = imports.gi;
+
+var PopupMenu = GObject.registerClass({
+    Properties: {
+        value: GObject.ParamSpec.string('value', 'Value', '',
+            GObject.ParamFlags.READWRITE, ''),
+    },
+}, class PopupMenu extends GObject.Object {
+    _init(button, groupDef, ItemClass, itemProp, itemExtraProps, props = {}) {
+        const entries = Object.entries(groupDef);
```
cosimoc, line 11: I wonder why you don't pass directly a Map to this method, instead of unpacking `groupDef`

src/popupMenu.js
```diff
@@ -0,0 +1,92 @@
...
+            columnSpacing: 6,
+            minChildrenPerLine,
+            maxChildrenPerLine,
+            visible: true,
+        });
+        this._button.popover = this._menu;
+
+        this._menu.add(this._choices);
+        entries.forEach(([, propValue]) => {
+            const choiceProps = {[itemProp]: propValue, visible: true};
```
cosimoc, line 38: One way to eliminate the need for the `propMap` back-table could be to store the enum value as well on this.

cosimoc (changes requested):

src/popupMenu.js
ptomato: In most cases the caller is going to want to define groupDef as a literal. There are no Map literals in JS, only Object literals.

src/framework/level1.js
ptomato: It is fixed, maybe you are looking at an old commit?

src/popupMenu.js
ptomato: Done.

src/framework/level1.js
cosimoc: Ah, you fixed it in the next commit.

src/framework/customModules/banner/hackableDynamic.js
```diff
@@ -86,8 +94,51 @@ var HackableDynamic = new Module.Class({
...
+        }
+        // eslint-disable-next-line no-new-func
+        this._decodefunc = new Function('letter', value);
+    },
+
+    // Adapted from https://gist.github.com/EvanHahn/2587465, public domain
     _processText(text) {
         const decomposedText = text.normalize('NFKD');
-        return TextFilters[this.textfilter](decomposedText);
+        let output = '';
```
cosimoc: I'd like to see this factored out to an utility function, since it's used in a few different places

cosimoc (changes requested):

src/framework/customModules/banner/hackableDynamic.js
```diff
@@ -36,6 +37,14 @@ var HackableDynamic = new Module.Class({
...
             GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
             'normal'),
+        rotation: GObject.ParamSpec.uint('rotation', 'Rotation',
+            'Number of positions to advance each letter in the string',
+            GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT_ONLY,
+            0, 25, 0),
+        decodefunc: GObject.ParamSpec.string('decodefunc', 'Decode Function',
+            'Source code of function to evaluate to decode the text',
+            GObject.ParamFlags.WRITABLE | GObject.ParamFlags.CONSTRUCT,
+            ''),
```
cosimoc, line 47: I commented on Slack that this would feel like a good use case for an interface implemented by all the modules that rotate text, so that the boilerplate does not have to be repeated over and over.
But that can be added later in a follow up cleanup, so I won't block the PR on it.

cosimoc (approved):

PR #19, opened 2018-10-10
T23984

tchx84:


tchx84:
Note that we can't merge this without https://github.com/endlessm/hack-toy-apps/pull/6 , otherwise this toolbox won't work.

src/hackyballs/level3.js
```diff
@@ -101,12 +101,47 @@ var HBLevel3 = GObject.registerClass({
...
+// Select Background
+/////////////////////
+
+backgroundImageIndex = ${this._model.backgroundImageIndex}
+
+/////////////////////
+// Object Parameters
+/////////////////////
+
+// Object 1
```
ptomato: Object 0

ptomato (changes requested):
The code all looks fine, but the control panel needs a bit of tightening up visually.

On the left side, the following things should all align vertically:
- The right edges of the labels on the left side of the sliders
- The left edges of the sliders
- The centers of the labels above the sliders
- The right edges of the sliders
- The left edges of the labels on the right side of the sliders

The "Freeze" switch has the opposite effect as intended, the balls freeze when it's turned off.

There should be more padding between the ball images on the right side, and the minus signs on the left edges of the sliders. Also, the minus signs should be Unicode minus signs (U+2212) rather than dashes.

A few comments about the wording, although I guess don't change this if it's what was in the specs. But "Attraction" seems a bit confusing when the term used in the code view is "Social force". Same goes for "Freeze" and "Use Physics". I also never much liked "Object 0", etc., maybe "Creature 0"?

tchx84:
> The code all looks fine, but the control panel needs a bit of tightening up visually.

I polished one thing though, I move the Gtk.Notebook and tabs defs to the .ui file, so the level1 code its cleaner there.

> On the left side, the following things should all align vertically:
>The right edges of the labels on the left side of the sliders
>The left edges of the sliders
>The centers of the labels above the sliders
>The right edges of the sliders
>The left edges of the labels on the right side of the sliders

Done. I tested with a physical ruler ;)

>The "Freeze" switch has the opposite effect as intended, the balls freeze when it's turned off.

Done. I switched the label to just "physics", so it matches the code better.

> There should be more padding between the ball images on the right side,

Done.

> and the minus signs on the left edges of the sliders. Also, the minus signs should be Unicode minus signs (U+2212) rather than dashes.

Done.

> A few comments about the wording, although I guess don't change this if it's what was in the specs. But "Attraction" seems a bit confusing when the term used in the code view is "Social force".

Done. Actually, this was wrong, since force > 0 means more repulsion , while force < 0 means attractions, which completely contradicts the UI, so I simply rename it to force, so its less confusing and matches the code.

> I also never much liked "Object 0", etc., maybe "Creature 0"?

Done. Noel calls them Balls, and since the game is still called hackyballs it makes sense we call them just balls for now.

tchx84:
From @erikos :

```
@martin you have my green light
so please, for hackyballs and the toolbox
and please make sure to build hackyballs master as well
so it gets out
( I just built it here, and nothing breaks at least)
```

PR #21, opened 2018-10-11
Fix popup menus and fonts

ptomato:
https://phabricator.endlessm.com/T23977

### DIFF
--- a/data/framework/level1.ui
+++ b/data/framework/level1.ui
@@ -115,25 +115,6 @@
               </packing>
             </child>
             <child>
-              <object class="GtkFontButton" id="fontButton">
-                <property name="visible">True</property>
-                <property name="can_focus">True</property>
-                <property name="receives_default">True</property>
-                <property name="valign">center</property>
-                <property name="hexpand">True</property>
-                <property name="font">Sans 12</property>
-                <property name="preview_text">Hack Everything</property>
-                <property name="use_font">True</property>
-                <property name="show_style">False</property>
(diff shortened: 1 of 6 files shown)


data/framework/level1.ui
```diff
@@ -305,6 +286,30 @@
...
+                  <object class="GtkCellRendererText" id="fontRenderer">
+                    <property name="cell_background_gdk">#000000000000</property>
+                    <property name="cell_background_rgba">rgba(0,0,0,0)</property>
+                    <property name="xpad">2</property>
+                    <property name="ypad">2</property>
+                    <property name="xalign">0</property>
+                    <property name="font">Normal</property>
+                    <property name="foreground_gdk">#000000000000</property>
+                    <property name="background_rgba">rgba(0,0,0,0)</property>
+                    <property name="foreground_rgba">rgba(0,0,0,0)</property>
```
cosimoc: I don't think these colors should be hardcoded here. This has the side-effect that when the font entry is selected from the combobox menu, it appears black-on-blue, which is hard to read.

cosimoc (changes requested):

data/framework/level1.ui
ptomato: Huh. This was completely autogenerated by Glade. Tomorrow I'll delete it and re-save the file and see what happens.

ptomato: OK, that worked. I guess I needed to disable the cell renderer properties instead of leaving them at their defaults

PR #23, opened 2018-10-12
framework: Don't relaunch the app if nothing changed

ptomato:
If nothing has changed since the first launch of the toolbox or the last
relaunch of the app, then don't relaunch the app; just flip back over.

Note, this is the simplest possible implementation and doesn't care
about the case where you change one of the parameters then change it
back to its original value.

https://phabricator.endlessm.com/T23991

### DIFF
--- a/src/framework/model.js
+++ b/src/framework/model.js
@@ -48,6 +48,8 @@ const _DEFAULTS = {
 
 var RaModel = GObject.registerClass({
     Properties: {
+        changed: GObject.ParamSpec.boolean('changed', 'Changed', '',
+            GObject.ParamFlags.READABLE, false),
         'logo-graphic': GObject.ParamSpec.string('logo-graphic', 'Logo Graphic', '',
             _propFlags, 'dinosaur'),
         'logo-color': GObject.ParamSpec.boxed('logo-color', 'Logo Color', '',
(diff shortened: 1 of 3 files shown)


src/framework/model.js
```diff
@@ -170,6 +190,7 @@ var RaModel = GObject.registerClass({
         if ('_borderWidth' in this && this._borderWidth === value)
             return;
         this._borderWidth = value;
+        this._changed = true;
```
cosimoc: Need to `this.notify('changed')` here too, no?

src/framework/model.js
```diff
@@ -94,6 +100,8 @@ var RaModel = GObject.registerClass({
             return;
         this._logoGraphic = value;
         this.notify('logo-graphic');
+        this._changed = true;
+        this.notify('changed');
```
cosimoc: Would be nice to have a `_setChanged()` function to avoid repeating these two lines over and over

src/window.js
```diff
@@ -9,15 +9,6 @@ const unlockState = {
     'com.endlessm.hackyballs': [false, false],
 };
 
-function _shouldEnableFlipBack(targetBusName) {
```
cosimoc, line 12: I assume that we don't need this special case anymore because the framework toolbox now controls whether it's enabled and it only gets enabled for dinosaurs?

cosimoc (changes requested):

cosimoc:
Also needs rebase.

src/window.js
ptomato: Correct, and in fact it needs to be disabled initially.

ptomato: We can add it back later if there's a toolbox where the app always needs to be restarted, or just use the toolbox's override of `bindWindow()` to set it.

src/framework/model.js
ptomato: I went with a different approach after all, without repeating this.

PR #25, opened 2018-10-12
T23984 (fixes)

tchx84:

### DIFF
--- a/src/hackyballs/toolbox.js
+++ b/src/hackyballs/toolbox.js
@@ -1,13 +1,17 @@
 /* exported HBToolbox */
 
 const {GObject} = imports.gi;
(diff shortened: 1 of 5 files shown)


src/hackyballs/toolbox.js
```diff
@@ -18,6 +18,7 @@ var HBToolbox = GObject.registerClass(class HBToolbox extends Toolbox {
         this.connect('reset', () => {
             this._model.reset();
         });
+        this.setBusy(false);
```
ptomato: This should probably be in the parent's constructor, I think we always want to start out not-busy.

ptomato (changes requested):
Only one minor comment.

PR #26, opened 2018-10-12
lockscreen: Fix events

ptomato:
The change to Gtk.EventBox meant that all clicks were being intercepted.
Remove the event box from the overlay after it is done fading out.

I'm still not sure why this was happening; the old Gtk.Frame would not
catch clicks even if its event mask included the correct event, while
Gtk.EventBox always intercepts clicks even when the event handler
returns Gdk.EVENT_PROPAGATE and its event mask was set to 0.

https://phabricator.endlessm.com/T24072

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -1,6 +1,8 @@
 /* exported Lockscreen */
 
-const {Gdk, GObject, Gtk} = imports.gi;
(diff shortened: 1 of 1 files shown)


tchx84 (approved):
It fixes the issue.

PR #28, opened 2018-10-13
T24015

cosimoc:
https://phabricator.endlessm.com/T24015


src/hackyballs/toolbox.js
```diff
@@ -26,6 +26,7 @@ var HBToolbox = GObject.registerClass(class HBToolbox extends Toolbox {
     }
 
     bindWindow(win) {
+        win.get_style_context().add_style_class('hackyballs');
```
ptomato: add_style_class -> add_class

ptomato (changes requested):
I haven't tested this, just looked at the code, but it looks fine, only spotted one mistake.

For the context of why I originally put the lock state in the window: then it was all in one place and we could later replace that with the global game state. However, I think this change is fine, as we don't have the lock state in global game state anyway yet, and it's not clear how it's going to look.

cosimoc:
Let's get this in for playtest.

PR #29, opened 2018-10-16
hackyballs: Use common widget for popup choices menu

ptomato:
This uses the common popup menu widget that we also use in the framework
toolbox, for consistency. The common widget visually selects the chosen
option, which would otherwise need to be solved separately in the
hackyballs toolbox code.

### DIFF
--- a/data/hackyballs/objectEditor.ui
+++ b/data/hackyballs/objectEditor.ui
@@ -47,20 +47,6 @@
     <property name="step_increment">15</property>
     <property name="page_increment">10</property>
   </object>
-  <object class="GtkPopover" id="menuSkin">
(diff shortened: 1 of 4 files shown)


tchx84:
@ptomato thanks for taking over this one. One comment:

It's not syncing with the initial state. e.g. open the app and the toolbar (for the first time), go to the "Balls 1" tab (the second one), and you will see that the button is not displaying the right icon.

I also see this error (probably related):
```
"JS ERROR: Gio.ResourceError: The resource at “/com/endlessm/HackToolbox/hackyballs/skins/undefined.png” does not exist"
```

tchx84 (changes requested):
One small regression.

PR #30, opened 2018-10-17
T23943

tchx84:
https://phabricator.endlessm.com/T23943

### DIFF
--- a/src/HackUnlock/model.js
+++ b/src/HackUnlock/model.js
@@ -1,0 +1,25 @@
+/* exported HUModelGlobal */
+
+const {GObject} = imports.gi;
+const {HBModelBase} = imports.hackyballs.model;
+
+const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
+
+var HUModelGlobal = GObject.registerClass({
+    Properties: {
+        amplitude: GObject.ParamSpec.double(
+            'amplitude', 'amplitude', '',
+            _propFlags, 0.0, 0.7, 0.2),
+        frequency: GObject.ParamSpec.double(
+            'frequency', 'frequency', '',
+            _propFlags, 0.0, 50.0, 20.0),
+        phase: GObject.ParamSpec.double(
+            'phase', 'phase', '',
+            _propFlags, -3.0, 3.0, 0.0),
+    },
+}, class HUModelGlobal extends HBModelBase {
+    _init(props = {}) {
+        super._init(props);
+        this.bindProperties('com.endlessm.HackUnlock', '/com/endlessm/HackUnlock', {});
+    }
+});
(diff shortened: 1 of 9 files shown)


src/HackUnlock/level1.js
```diff
@@ -0,0 +1,20 @@
+/* exported HULevel1 */
```
cosimoc: I don't think the HackUnlock will ever get more than one level, so this feels an unnecessary abstraction at that point, and could be merged back with the controlpanel itself.

src/HackUnlock/model.js
```diff
@@ -0,0 +1,25 @@
+/* exported HUModelGlobal */
+
+const {GObject} = imports.gi;
+const {HBModelBase} = imports.hackyballs.model;
```
cosimoc, line 4: It feels to me that the first step here is then to lift the `HBModelBase` class from the hackyballs directory and rename it to be more generic, e.g. `ToyAppModel`.

src/HackUnlock/model.js
```diff
@@ -0,0 +1,25 @@
...
+            'frequency', 'frequency', '',
+            _propFlags, 5.0, 30.0, 20.0),
+        phase: GObject.ParamSpec.double(
+            'phase', 'phase', '',
+            _propFlags, -2.0, 2.0, 0.0),
+    },
+}, class HUModelGlobal extends HBModelBase {
+    _init(props = {}) {
+        super._init(props);
+        this.bindProperties('com.endlessm.HackUnlock', '/com/endlessm/HackUnlock', {});
```
cosimoc, line 23: I'm a bit confused, don't you need to pass the property map here? There should be at least some no?

cosimoc (changes requested):

src/HackUnlock/level1.js
tchx84: I kind of like the consistency, but this makes sense.

src/HackUnlock/model.js
tchx84: All this piece of code should be removed soon, since it should be replace with @xjuan 's new clippy API. I rather just remove it all together then, than refactoring this more.

src/HackUnlock/model.js
tchx84: The base class assume that if no/empty map is provided then the properties names are the same locally and remotely, but same as my comment above, ill should be able to remove all of this with new Clippy API.

src/HackUnlock/model.js
cosimoc: Sounds good

src/HackUnlock/model.js
cosimoc: Oh, I see. I'll wait for that to land then before making any further refactoring suggestions :-)

PR #32, opened 2018-10-22
T24156

tchx84:
https://phabricator.endlessm.com/T24156

### DIFF
--- a/src/HackUnlock/toolbox.js
+++ b/src/HackUnlock/toolbox.js
@@ -13,7 +13,6 @@ var HUToolbox = GObject.registerClass(class HUToolbox extends Toolbox {
     _init(props = {}) {
         props.title = _('Hack Modules');
         super._init(props);
-        this._unlockState = [false];
         this._model = new HUModelGlobal();
         this._controlPanel = new HUControlPanel({visible: true});
         this._controlPanel.bindModel(this._model);
@@ -27,6 +26,7 @@ var HUToolbox = GObject.registerClass(class HUToolbox extends Toolbox {
 
     bindWindow(win) {
         win.get_style_context().add_class('HackUnlock');
+        win.lockscreen.locked = false;
         this._win = win;
     }
 });
--- a/src/hackyballs/controlpanel.js
+++ b/src/hackyballs/controlpanel.js
@@ -20,7 +20,6 @@ var HBControlPanel = GObject.registerClass({
         this._level2 = new HBLevel2({visible: true});
         this._level2lock = new Lockscreen({visible: false});
 
-
         this._level2lock.add(this._level2);
         this._panelLevel1.add(this._level1);
         this._panelLevel2.add(this._level2lock);
@@ -33,17 +32,8 @@ var HBControlPanel = GObject.registerClass({
         this._model = model;
     }
 
-    _onUnlockStateChanged(win) {
-        const [, l2] = win.getUnlockState();
-        if (l2)
-            this._level2lock.locked = false;
-    }
-
     bindWindow(win) {
-        win.connect('unlock-state-changed', this._onUnlockStateChanged.bind(this));
-        this._onUnlockStateChanged(win);
-
-        // FIXME For playtest only. Remove this later.
-        this._level2lock.connect('overlay-clicked', () => win.unlock());
+        win.lockscreen.key = 'item.key.fizzics.1';
+        this._level2lock.key = 'item.key.fizzics.2';
     }
 });
--- a/src/locksManager.js
+++ b/src/locksManager.js
@@ -1,0 +1,73 @@
+/* exported LocksManager */
+
+const {Gio, GLib, GObject} = imports.gi;
+
+const BusName = 'com.endlessm.GameStateService';
+const BusPath = '/com/endlessm/GameStateService';
+const BusIface = `
+    <node>
+      <interface name='com.endlessm.GameStateService'>
+        <method name='Get'>
+          <arg type='s' name='key' direction='in'/>
+          <arg type='v' name='value' direction='out'/>
+        </method>
+        <method name='Set'>
+          <arg type='s' name='key' direction='in'/>
+          <arg type='v' name='value' direction='in'/>
+        </method>
+        <signal name='changed'>
+          <arg type='s' name='key'/>
+          <arg type='v' name='value'/>
+        </signal>
+      </interface>
+    </node>
+`;
+
+var LocksManager = GObject.registerClass({
+    Signals: {
+        changed: {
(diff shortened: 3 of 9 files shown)


src/locksManager.js
```diff
@@ -0,0 +1,71 @@
...
+var LocksManager = GObject.registerClass({
+    Signals: {
+        changed: {},
+    },
+}, class LocksManager extends GObject.Object {
+    _init(props = {}) {
+        super._init(props);
+        const Proxy = Gio.DBusProxy.makeProxyWrapper(BusIface);
+        this._proxy = new Proxy(Gio.DBus.session, BusName, BusPath);
+        this._proxy.connect('g-signal', this._onChanged.bind(this));
```
cosimoc, line 37: How can this do anything, given that the DBus interface does not have any signals in it?

src/locksManager.js
```diff
@@ -0,0 +1,71 @@
...
+            return false;
+        }
+    }
+
+    setUnlocked(key) {
+        const dict = GLib.Variant.new_dict_entry(
+            new GLib.Variant('s', 'used'),
+            new GLib.Variant('b', true));
+        const builder = new GLib.VariantBuilder(new GLib.VariantType('a{sb}'));
+        builder.add_value(dict);
```
cosimoc: Yeah, let's try if we can use a simpler syntax like @ptomato suggested.

src/lockscreen.js
```diff
@@ -42,6 +49,39 @@ var Lockscreen = GObject.registerClass({
...
+    }
+
+    set key(key) {
+        if ('_key' in this && this._key === key)
+            return;
+        this._key = key;
+        this._updateLockState();
+    }
+
+    _onChanged() {
```
cosimoc, line 67: Looking at how this signal is used, it looks to me that you'd want a signal with details, with the detail being the key itself. So here the lockscreen could connect to `changed::key` and avoid being waken up for any other change.

src/lockscreen.js
```diff
@@ -42,6 +49,39 @@ var Lockscreen = GObject.registerClass({
...
+            this.locked = false;
+    }
+
+    _onClicked() {
+        if (!this._key)
+            return;
+        if (!this._manager.hasKey(this._key))
+            return;
+        this.locked = false;
+        this._manager.setUnlocked(this._key);
```
cosimoc, line 86: Wouldn't it be more practical to have the `locked` setter take care of also calling `setUnlocked()`?

src/window.js
```diff
@@ -82,30 +69,14 @@ var ToolboxWindow = GObject.registerClass({
...
-    unlock() {
-        const unlockState = this.getUnlockState();
-        const index = unlockState.findIndex(state => !state);
-        this.setUnlockState(index, true);
-        this.emit('unlock-state-changed');
-
-        // Update state of the common lockscreen in this window
-        this._lockscreen.locked = !this.getUnlockState()[0];
+    getMainLockscreen() {
+        return this._lockscreen;
```
cosimoc, line 73: I find that it would be more consistent to have this be a simple getter, since that's what we do for the toolbox.

src/hackyballs/controlpanel.js
```diff
@@ -33,17 +32,8 @@ var HBControlPanel = GObject.registerClass({
...
-    }
-
     bindWindow(win) {
-        win.connect('unlock-state-changed', this._onUnlockStateChanged.bind(this));
-        this._onUnlockStateChanged(win);
-
-        // FIXME For playtest only. Remove this later.
-        this._level2lock.connect('overlay-clicked', () => win.unlock());
+        win.getMainLockscreen().key = 'item.key.fizzics.1';
+        this._level2lock.key = 'item.key.fizzics.2';
```
cosimoc, line 37: Could probably be set at construction, no? (also for other lockscreens)

src/HackUnlock/toolbox.js
```diff
@@ -27,6 +26,7 @@ var HUToolbox = GObject.registerClass(class HUToolbox extends Toolbox {
 
     bindWindow(win) {
         win.get_style_context().add_class('HackUnlock');
+        win.getMainLockscreen().locked = false;
```
cosimoc: Hmm... Previously this had a `false` unlock state, which corresponds to a `true` locked state, no?

cosimoc (changes requested):

src/locksManager.js
tchx84: good catch, no idea why it is actually working! But I will add it to the iface.

src/lockscreen.js
tchx84: it's a bit confusing but, there is only 1 specific scenario where we should set it to "unlocked" in the GameStateService, and that is when the player has used the key.

In other scenarios, e.g. when the locked has been unlocked in a previous session, then setting the lockscreen.locked to false should only affect UI, not the GameStateService.

src/hackyballs/controlpanel.js
tchx84: A couple reasons for doing it like this: a) we can't do it for the window's lockscreen, b) like it happens with the framework lockscreens, some times we need to define some rules before the locks can actually be unlocked (so its simple assign keys only after we have all the lockscreens available), and c) I think is just easier to read to have all those strings in one place.

src/HackUnlock/toolbox.js
tchx84: i will need to confirm with @llopis , but I think this one is never locked.. but if it is locked, then the quest for the Unlock/Welcome will need to provide a key, and we will need to use it here.

src/lockscreen.js
```diff
@@ -28,6 +31,10 @@ var Lockscreen = GObject.registerClass({
             this.emit('overlay-clicked');
             return Gdk.EVENT_PROPAGATE;
         });
+
+        this._manager = new LocksManager();
+        this._manager_changed_id = null;
```
cosimoc: We use camel case for variables in JS; also this is a signal connection, for which it's more typical to use 0 as the initialization value.

src/lockscreen.js
```diff
@@ -42,6 +49,43 @@ var Lockscreen = GObject.registerClass({
...
 
+    get key() {
+        return this._key;
+    }
+
+    set key(key) {
+        if ('_key' in this && this._key === key)
+            return;
+        if (this._manager_changed_id !== null)
+            GLib.source_remove(this._manager_changed_id);
```
cosimoc: This is not a source, it's a signal connection; you have to use `this._manager.disconnect()`

cosimoc (changes requested):

src/lockscreen.js
tchx84: Time for a bicycling.

PR #33, opened 2018-10-22
OperatingSystemApp: implement wobbly window controls

xjuan:
https://phabricator.endlessm.com/T24045

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -43,6 +43,13 @@ toolbox.dark {
     background-color: #33393b;  /* Adwaita-dark background color */
 }
 
+/* Toolbox HeaderBar */
+frame.toolbox > border,
+frame.toolbox > toolbox,
+frame.toolbox > toolbox > headerbar {
+    border-radius: 8px 8px 0px 0px;
+}
+
 /* Temporarily make the buttons smaller until we figure out how to layout */
 button.radio {
     font-size: 10px;
--- a/data/application.css
+++ b/data/application.css
@@ -69,6 +69,7 @@ popover.error {
     background-color: rgba(0, 0, 0, 0.6);
 }
 
+.OperatingSystemApp.toolbox-surrounding-window,
 .HackUnlock.toolbox-surrounding-window,
 .hackyballs.toolbox-surrounding-window {
     background-color: rgba(0, 0, 0, 0);
--- a/src/OperatingSystemApp/controlpanel.js
+++ b/src/OperatingSystemApp/controlpanel.js
@@ -1,10 +1,88 @@
-/* exported OSControlPanel */
+/* exported OSControlPanel, OSModel */
 
-const {GObject, Gtk} = imports.gi;
+const {Gio, GObject, Gtk} = imports.gi;
+const {Codeview} = imports.codeview;
+const {Lockscreen} = imports.lockscreen;
+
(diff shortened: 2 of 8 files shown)


data/application.css
```diff
@@ -43,6 +43,13 @@ toolbox.dark {
     background-color: #33393b;  /* Adwaita-dark background color */
 }
 
+/* Toolbox HeaderBar */
+frame.toolbox > border,
+frame.toolbox > toolbox,
+frame.toolbox > toolbox > headerbar {
```
cosimoc, line 49: Do you need all three selectors? Why isn't the headerbar only not enough?

src/OperatingSystemApp/controlpanel.js
```diff
@@ -1,10 +1,88 @@
...
+        this._keys = this._settings.settings_schema.list_keys();
+
+        this.bindSetting('wobbly-effect', this, 'wobblyEffect');
+        this.bindSetting('wobbly-object-movement-range', this, 'wobblyObjectMovementRange');
+        this.bindSetting('wobbly-slowdown-factor', this, 'wobblySlowdownFactor');
+        this.bindSetting('wobbly-spring-friction', this, 'wobblySpringFriction');
+        this.bindSetting('wobbly-spring-k', this, 'wobblySpringK');
+    }
+
+    bindSetting(setting, object, property) {
```
cosimoc: This can be private (start with an underscore)

cosimoc (changes requested):

src/OperatingSystemApp/controlpanel.js
xjuan: yeah I forgot to make it private (I was using outside the class before)

data/application.css
xjuan: because the parent containers also draw backgrounds or borders so you need all three of them

src/OperatingSystemApp/toolbox.js
```diff
@@ -14,6 +14,7 @@ var OSToolbox = GObject.registerClass(class OSToolbox extends Toolbox {
 
     bindWindow(win) {
         void (this, win);
```
cosimoc, line 16: I think you can remove this line now

src/OperatingSystemApp/controlpanel.js
```diff
@@ -1,10 +1,88 @@
...
+        this._keys = this._settings.settings_schema.list_keys();
+
+        this._bindSetting('wobbly-effect', 'wobblyEffect');
+        this._bindSetting('wobbly-object-movement-range', 'wobblyObjectMovementRange');
+        this._bindSetting('wobbly-slowdown-factor', 'wobblySlowdownFactor');
+        this._bindSetting('wobbly-spring-friction', 'wobblySpringFriction');
+        this._bindSetting('wobbly-spring-k', 'wobblySpringK');
+    }
+
+    bindSetting(setting, property) {
```
cosimoc: Looks like the calls were renamed, but not the function itself...

cosimoc (changes requested):

src/OperatingSystemApp/toolbox.js
xjuan: unfortunately I can not, because it complains the function should use the this pointer :S

PR #38, opened 2018-10-26
OperatingSystemApp: Add basic cursors support

xjuan:
Add support for cursor theme, size and speed settings

https://phabricator.endlessm.com/T24194


erikos:
Thanks Juan Pablo!

Some observations:
- the size of those cursors is small when you change them, I think they get effective starting from size 80, maybe we should change the size automatically?
- the cone cursor gets a "pile" when over the taskbar
- the "pile" cursor gets a cone when over the taskbar
- fire->pile
- disco->fire
- unicorn->disco
.... is that intentional?

tchx84:
A couple comments:
* This is not really mandatory, but wouldn't it be more manageable if we broke the control panel in levels like we do in other toolboxes?
* Enabling the codeview !
* Lockscreen have a "key" property, so you can specify the GSS key for it.. e.g. "item.key.OperatingSystemApp.0". Would be best to do it in the control panel bindWindow function (because you have to access the window lockscreen as well).

Really cool stuff!

xjuan:
PR updated, need review again!

src/OperatingSystemApp/osmodel.js
```diff
@@ -0,0 +1,49 @@
...
+        var keys = customSettings ? customKeys : null;
+
+        if (keys && !keys.includes(setting))
+            return;
+
+        /* Keep track of all settings objects used */
+        if (this._settingsObjects.indexOf(settings) < 0)
+            this._settingsObjects.push(settings);
+
+        /* Also keep track of binded settings */
```
cosimoc: Typo: I think "bound" is the word to use here.

src/OperatingSystemApp/osmodel.js
```diff
@@ -0,0 +1,49 @@
...
+            path: props.schemaPath,
+        });
+
+        delete props.schemaId;
+        delete props.schemaPath;
+
+        super._init(props);
+
+        this._settingsObjects = [this.settings];
+        this.keys = this.settings.settings_schema.list_keys();
```
cosimoc: Does not look that this is used.

src/OperatingSystemApp/osmodel.js
```diff
@@ -0,0 +1,49 @@
+/* exported OSModel */
+const {Gio, GObject} = imports.gi;
+
+var OSModel = GObject.registerClass({
+}, class OSModel extends GObject.Object {
+    _init(props) {
+        this.settings = new Gio.Settings({
+            schemaId: props.schemaId,
+            path: props.schemaPath,
+        });
```
cosimoc: It feels odd that there is both the "regular" settings object and those that can be provided to `bindSetting()`. You could simply rely on `bindSetting()` instead, and require that a settings object is passed in.

src/OperatingSystemApp/osmodel.js
```diff
@@ -0,0 +1,49 @@
...
+        delete props.schemaId;
+        delete props.schemaPath;
+
+        super._init(props);
+
+        this._settingsObjects = [this.settings];
+        this.keys = this.settings.settings_schema.list_keys();
+    }
+
+    bindSetting(setting, property, customSettings, customKeys) {
```
cosimoc: It does not look like `customKeys` are ever passed in

src/OperatingSystemApp/osmodel.js
```diff
@@ -0,0 +1,49 @@
...
+        var settings = customSettings || this.settings;
+        var keys = customSettings ? customKeys : null;
```
cosimoc: Why not list the keys on the custom settings directly?

src/OperatingSystemApp/oswobblymodel.js
```diff
@@ -0,0 +1,39 @@
+/* exported OSWobblyModel */
+
+const {GObject} = imports.gi;
+const {OSModel} = imports.OperatingSystemApp.osmodel;
+
+var _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
```
cosimoc: This can be `const`

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+
+        this.bindSetting('cursor-size', 'size');
+        this.bindSetting('cursor-theme', 'theme');
+        this.bindSetting('speed', 'speed', this._cursorSettings);
+    }
+
+    reset() {
+        this.theme = 'cursor-default';
+        this.settings.reset('cursor-size');
+        this._cursorSettings.reset('speed');
```
cosimoc: I think it would be best to chain up to parent here, which would do the same.

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
+/* exported OSCursorModel */
+
+const {Gio, GLib, GObject} = imports.gi;
+const {OSModel} = imports.OperatingSystemApp.osmodel;
+
+var _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
```
cosimoc: `const` could be used here and in a few other places in this file

src/OperatingSystemApp/controlpanel.js
```diff
@@ -1,88 +1,93 @@
...
+
+    reset () {
+        this._cursor.reset();
+        this._wobbly.reset();
+    }
+
+    _refreshRadioGroup() {
+        var group = this._cursorRadiobutton.get_group();
+
+        for (var i = 0, n = group.length; i < n; i++) {
```
cosimoc: You can use the more idiomatic `for (var button of group)`

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+
+        super._init(props);
+
+        /* Get resource mtime to know if we need to update icons or not */
+        try {
+            var gresource = Gio.File.new_for_path(
+                '/app/share/com.endlessm.HackToolbox/com.endlessm.HackToolbox.data.gresource');
+            var gresource_info = gresource.query_info('time::modified', 0, null);
+            this._gresource_mtime = gresource_info.get_attribute_uint64('time::modified');
+        } catch (e) {
```
cosimoc: This should at least initialize `this._gresource_mtime = 0`, but also log the error

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+
+        this._theme = cursorThemes.indexOf(cursor_theme) < 0 ? 'cursor-default' : cursor_theme;
+
+        /* Build user custom cursors directory path */
+        var dir = Gio.File.new_for_path(GLib.build_pathv('/',
+            [GLib.get_home_dir(), '.icons', this._theme, 'cursors']));
+        var cursor_file = dir.get_child('left_ptr');
+        var cursor_info, cursor_mtime;
+
+        /* If cursor exits and its older than the resource file,
```
cosimoc: `exists` and `it's older`. But isn't it the other way around? This block will return and notify when the mtime of the cursor is higher than the gresource, which means that it's newer.

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+         * then all we have to do is notify!
+         */
+        if (cursor_file.query_exists(null) &&
+            (cursor_info = cursor_file.query_info('time::modified', 0, null)) &&
+            (cursor_mtime = cursor_info.get_attribute_uint64('time::modified')) &&
+            cursor_mtime >= this._gresource_mtime) {
```
cosimoc: I am not sure this mtime check really does what you intend it to do. The intention as I understand it is to make it so that a newer cursor is copied over only if we update the gresource asset. However, the timestamp of the gresource is managed by Flatpak/ostree and is always set to zero. (You will see it's from 12/31/1969 if you do `ls -l` on a deployed flatpak directory in the terminal.)

However, I don't think that it's a big deal to always copy the asset over when something changes, so perhaps we can just do that.

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+            this.notify('theme');
+            return;
+        }
+
+        /* Ensure cursor theme directory exists */
+        try {
+            dir.make_directory_with_parents(null);
+        } catch (e) {
```
cosimoc, line 103: I'd like to print the error here if it's not `G_IO_ERROR_EXISTS`

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+        var dir = Gio.File.new_for_path(GLib.build_pathv('/',
+            [GLib.get_home_dir(), '.icons', this._theme, 'cursors']));
+        var cursor_file = dir.get_child('left_ptr');
+        var cursor_info, cursor_mtime;
+
+        /* If cursor exits and its older than the resource file,
+         * then all we have to do is notify!
+         */
+        if (cursor_file.query_exists(null) &&
+            (cursor_info = cursor_file.query_info('time::modified', 0, null)) &&
```
cosimoc: `query_info()` can throw an exception

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,114 @@
...
+                this._theme}.xmc`);
+
+        /* Copy cursor to ~/.icons/${cursor-shape}/cursors/left_ptr */
+        cursor.copy_async(cursor_file,
+            Gio.FileCopyFlags.OVERWRITE | Gio.FileCopyFlags.TARGET_DEFAULT_PERMS,
+            GLib.PRIORITY_DEFAULT,
+            null,
+            null,
+            (source, res) => {
+                source.copy_finish(res);
```
cosimoc: `copy_finish()` can throw an exception

cosimoc (changes requested):

src/OperatingSystemApp/toolbox.js
```diff
@@ -10,10 +10,16 @@ var OSToolbox = GObject.registerClass(class OSToolbox extends Toolbox {
...
         this.show_all();
+
+        this.connect('reset', () => {
+            this._controlPanel.reset();
+        });
     }
 
     bindWindow(win) {
-        void (this, win);
+        win.lockscreen.key = 'item.key.OperatingSystemApp.1';
```
llopis, line 20: I'm changing the quests to give out this key name, so let's make sure this name doesn't change.

src/OperatingSystemApp/osmodel.js
xjuan: Good idea

src/OperatingSystemApp/oscursormodel.js
xjuan: The parent method would reset everything to the gsettings default, but we do not want that becaus e the actual default cursor only goes up to size 96, and I made a version that goes up until 256 like the rest

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,110 @@
...
+        var dir = Gio.File.new_for_path(GLib.build_pathv('/',
+            [GLib.get_home_dir(), '.icons', this._theme, 'cursors']));
+        var cursor_file = dir.get_child('left_ptr');
+
+        /* Ensure cursor theme directory exists */
+        try {
+            dir.make_directory_with_parents(null);
+        } catch (e) {
+            if (e.code !== Gio.IOErrorEnum.EXISTS)
+                print(e.message);
```
cosimoc: Use `e.matches()` and `logError(e)`.

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,110 @@
...
+            GLib.PRIORITY_DEFAULT,
+            null,
+            null,
+            (source, res) => {
+                try {
+                    source.copy_finish(res);
+                    this._copiedCursors.push(this._theme);
+                    this.notify('theme');
+                } catch (e) {
+                    print(e.message);
```
cosimoc: Use `logError()`

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+
+        /* Build cursor resource path */
+        var cursor = Gio.File.new_for_uri(
+            `resource:///com/endlessm/HackToolbox/OperatingSystemApp/cursors/${
+                this._theme}.xmc`);
+
+        /* Build user custom cursors directory path */
+        var dir = Gio.File.new_for_path(GLib.build_pathv('/',
```
cosimoc: Can you use `GLib.build_filenamev()` and remove the path separator here?

src/OperatingSystemApp/osmodel.js
cosimoc: We discussed in person with @xjuan that this still needs to be addressed.

src/OperatingSystemApp/oscursormodel.js
```diff
@@ -0,0 +1,110 @@
...
+            [GLib.get_home_dir(), '.icons', this._theme, 'cursors']));
```
cosimoc, line 97: Try to see if `~/.local/share/icons` works too (i.e. `GLib.get_user_data_dir()/share/icons`)

src/OperatingSystemApp/oscursormodel.js
cosimoc: We discussed in person that this can be done by chaining up to parent and then re-setting our own tweaked default.

cosimoc (changes requested):

src/OperatingSystemApp/oscursormodel.js
cosimoc: (It doesn't. libXCursor which is what is used to load cursor themes does not look in there, unfortunately.)

PR #40, opened 2018-10-26
T23926

tchx84:
https://phabricator.endlessm.com/T23926


tchx84:
Looks like this will need to be rebased on top of https://github.com/endlessm/hack-toolbox-app/pull/41

xjuan:
CR Pass

PR #42, opened 2018-10-27
Make Hackdex Chapter 1 hackable

ptomato:
https://phabricator.endlessm.com/T24204

### DIFF
--- a/data/framework/app-descriptions/Hackdex_chapter_one.yaml
+++ b/data/framework/app-descriptions/Hackdex_chapter_one.yaml
@@ -1,0 +1,130 @@
+---
+# Keep in sync with the app.yaml from com.endlessm.Hackdex_chapter_one
+vars:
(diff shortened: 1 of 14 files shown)


data/framework/app-descriptions/Hackdex_chapter_one.yaml
```diff
@@ -0,0 +1,130 @@
+---
+# Keep in sync with the app.yaml from com.endlessm.Hackdex_chapter_one
```
cosimoc, line 2: Yuck... the toolbox already has `host` file system permissions. It would not be crazy to give it explicit permission to the flatpak installation location and have it fetch from there.
Really though, this should probably be an SDK DBus API...
Not going to block the PR on it, but it may cause big headaches if the toolbox and the app are not updated in lockstep, so I'd love to find a more robust fix.

PR #46, opened 2018-10-31
Hackdex Chapter 1 "corruption" quest

ptomato:
This is the toolbox half of the implementation of this quest. (It won't work without the corresponding changes in the Hackdex Chapter 1 app.)

https://phabricator.endlessm.com/T24258

### DIFF
--- a/src/framework/defaults.js
+++ b/src/framework/defaults.js
@@ -2,6 +2,8 @@
 
 const {Gdk} = imports.gi;
 
+const GameState = imports.gameState;
+
 const COLORS = Symbol('colors');
 
 const DEFAULTS = {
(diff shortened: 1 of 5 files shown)


src/framework/defaults.js
```diff
@@ -65,6 +67,30 @@ var Defaults = class Defaults {
...
+    }
+
+    _applyQuestOverridesSync() {
+        if (this._busName === 'com.endlessm.Hackdex_chapter_one') {
+            // Override for the "corrupted hackdex chapter 1" quest
+            const gameState = GameState.getDefault();
+            const key = 'app.com_endlessm_Hackdex_chapter_one.corruption';
+            let corruption;
+            try {
+                corruption = gameState.GetSync(key);
```
tchx84, line 80: curious: why not using `await gameState.Get(key)` here? Like you do  in the model.

tchx84 (approved):

PR #47, opened 2018-10-31
framework: Work around limitation in autobahn include paths

ptomato:
This unbreaks flipping the dinosaurs app. I did not realize that
autobahn only took one include path.

https://phabricator.endlessm.com/T24295


cosimoc:
Should we also fix this limitation in autobahn?

ptomato:
Already did... https://github.com/endlessm/eos-knowledge-lib/pull/1690

PR #51, opened 2018-11-02
Background asset and alpha channels

ptomato:
https://phabricator.endlessm.com/T24321


data/application.css
```diff
@@ -36,7 +36,13 @@ widget.lockscreen.locked {
 }
 
 toolbox {
-    background-color: #33393b;  /* Adwaita-dark background color */
+    background-image: url('resource:///com/endlessm/HackToolbox/hacktoolbox/background.png'),
+                      radial-gradient(at -47% 158%,
+                                      rgba(197, 163, 93, 0.81) 0%,
+                                      rgba(208, 95, 52, 0.80) 14%,
+                                      rgba(32, 69, 108, 0.63) 66%,
+                                      #1C3753 100%),
```
cosimoc, line 44: I thought radial gradients did not work?

cosimoc (changes requested):

data/application.css
ptomato: That was my bad :smile: The GTK documentation says they do not work, but I used them myself elsewhere in the CSS file and they worked fine. I was just missing the `at` keyword

PR #52, opened 2018-11-02
Hackdex Chapter 1 should use $accent-light-color for chapter title

masimka:


src/framework/gen.js
```diff
@@ -461,7 +461,7 @@ $body-font: ${model.font};
 
 .bannerHead {
     font-size: ${model.font_size * 2.5}px;
-    color: $primary-light-color;
+    color: $accent-light-color;
```
ptomato, line 465: It has to be added to the variable list at the top of this generated CSS, as well.

ptomato (changes requested):

src/framework/gen.js
masimka: $accent-light-color is already defined on line 142, or should that be added to another line?

masimka: Done

ptomato (approved):

PR #59, opened 2018-11-06
codeview: Add keypress sounds

ptomato:
This adds sounds when typing space, equals, slash, and enter.

https://phabricator.endlessm.com/T24355


src/codeview.js
```diff
@@ -158,6 +164,27 @@ var Codeview = GObject.registerClass({
         this._helpMessage.popup();
     }
 
+    static _onInsertText(buffer, location, text) {
+        const sound = SoundServer.getDefault();
+        [...text].forEach(letter => {
```
cosimoc: I feel that you could simplify this by having a dictionary mapping letter -> sound name. Then you can look at `dictionary[letter]` and play that if it exists, without having to enumerate every case in the switch.

cosimoc (changes requested):

PR #64, opened 2018-11-07
Live update sounds

ptomato:
https://phabricator.endlessm.com/T24355

### DIFF
--- a/src/Fizzics/level2.js
+++ b/src/Fizzics/level2.js
@@ -39,7 +39,7 @@ var FizzicsLevel2 = GObject.registerClass({
 
         const scope = {};
         this._getModelProps().forEach(pspec => {
-            scope[this._toScopeName(pspec.get_name())] = null;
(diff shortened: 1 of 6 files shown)


src/Fizzics/level2.js
```diff
@@ -70,8 +74,17 @@ var FizzicsLevel2 = GObject.registerClass({
...
         this._codeview.setCompileResults([]);
 
+        // Block the normal notify handler that updates the code view, since we
+        // are propagating updates from the codeview to the GUI. Instead,
+        // connect a temporary handler that lets us know if anything actually
+        // did change.
         GObject.signal_handler_block(this._model, this._notifyHandler);
 
+        let guiUpdated = false;
+        this._tempHandler = this._model.connect('notify', () => {
```
cosimoc: You don't need to store this in `this`, it can be a local variable. (Same below.)

cosimoc (changes requested):
I'll just fix this locally and merge it, so that we can test it on tonight's image.

PR #66, opened 2018-11-09
T24375

tchx84:
https://phabricator.endlessm.com/T24375


src/lockscreen.js
```diff
@@ -56,24 +59,42 @@ var Lockscreen = GObject.registerClass({
...
     }
 
-    _updateLockState() {
+    _updateLockStateWithKey() {
         if (!this._key)
             return;
-        if (!this._manager.isUnlocked(this._key))
+        if (this._manager.hasKey(this._key)) {
+
+            /* TODO update the lock-screen graphics with can-be-opened image */
```
manuq, line 90: It would be great to add at least a mock graphic for now, for the quests walkthrough.

PR #67, opened 2018-11-10
OperatingSystemApp: generate index.theme file for cursor themes

cosimoc:
So that we can specify we inherit from Adwaita.

https://phabricator.endlessm.com/T24413


xjuan:
CR Pass

PR #68, opened 2018-11-13
T24465

tchx84:
https://phabricator.endlessm.com/T24465


ptomato:
Works OK for me!

For the future I would appreciate if you made the changes such as the semicolons and the spinner padding in the other toolboxes as well, so everything remains consistent, but this is fine, let's get it merged.

PR #69, opened 2018-11-13
Semicolons and spacing

ptomato:


data/framework/level2.ui
```diff
@@ -103,6 +103,7 @@
                   <object class="GtkLabel">
                     <property name="visible">True</property>
                     <property name="can_focus">False</property>
+                    <property name="margin_top">5</property>
```
tchx84, line 106: This is correct, but to be even more consistent, we should apply this to "hover" and "click" options as well.

PR #73, opened 2018-11-14
T24492

tchx84:
https://phabricator.endlessm.com/T24492

### DIFF
--- a/src/Fizzics/toolbox.js
+++ b/src/Fizzics/toolbox.js
@@ -1,6 +1,6 @@
 /* exported FizzicsToolbox */
 
-const {GObject} = imports.gi;
+const {Gio, GLib, GObject} = imports.gi;
 const Gettext = imports.gettext;
 
 const {Toolbox} = imports.toolbox;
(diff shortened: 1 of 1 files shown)


src/Fizzics/toolbox.js
```diff
@@ -20,7 +20,13 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
         this.show_all();
 
         this.connect('reset', () => {
-            this._model.reset();
+            const data = new GLib.Variant('a{sv}');
+            const action = Gio.DBusActionGroup.get(
+                Gio.bus_get_sync(Gio.BusType.SESSION, null),
```
ptomato: For a sync action, use `Gio.DBus.session`. Otherwise use an async operation.

src/Fizzics/toolbox.js
```diff
@@ -20,7 +20,13 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
...
+                'com.endlessm.Fizzics',
+                '/com/endlessm/Fizzics',
+            );
+            action.activate_action_full('reset', null, data);
```
ptomato: Use `action.activate_action('reset')`, it looks like the platform data is not necessary here.

src/Fizzics/toolbox.js
```diff
@@ -20,7 +20,13 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
         this.show_all();
 
         this.connect('reset', () => {
-            this._model.reset();
+            const data = new GLib.Variant('a{sv}');
+            const action = Gio.DBusActionGroup.get(
```
ptomato: I think an easier way to call an action remotely might be this:
```js
Gio.DBus.session.call_sync('com.endlessm.Fizzics',
    '/com/endlessm/Fizzics',
    'org.gtk.Actions', 'Activate',
    null, Gio.DBusCallFlags.NONE, -1, null);
```
or the async variant thereof (remember that `Gio.DBus.session` hides a sync operation)

ptomato (changes requested):

ptomato (approved):

PR #74, opened 2018-11-14
T24498

tchx84:
https://phabricator.endlessm.com/T24498


src/toolbox.js
```diff
@@ -58,15 +58,23 @@ var Toolbox = GObject.registerClass({
...
 
     get title() {
         return this._title;
     }
 
+    get resetButton() {
+        return this._buttonReset;
+    }
+
+    get buttonMinimize() {
```
cosimoc, line 74: I find it confusing that one button is called `resetButton` but the other one is called `buttonMinimize` and not `minimizeButton`.
Either make both start with `button` or not :)

cosimoc (changes requested):
Looks good with the exception of this comment.

PR #75, opened 2018-11-15
T24511

tchx84:
https://phabricator.endlessm.com/T24511


tchx84:
We are putting this on wait, until we have more concretes ideas and strings.

PR #76, opened 2018-11-15
T24348

tchx84:
https://phabricator.endlessm.com/T24348

### DIFF
--- a/src/gameState.js
+++ b/src/gameState.js
@@ -52,6 +52,16 @@ var getDefault = (function () {
                     });
                 });
             };
+            // Helpers
+            defaultGameStateProxy.getDictValueSync = function(key, property, defaultValue) {
+                try {
+                    const [variant] = this.GetSync(key);
+                    const dict = variant.deep_unpack();
+                    return dict[property].deep_unpack();
+                } catch (e) {
+                    return defaultValue;
+                }
+            };
         }
         return defaultGameStateProxy;
     };
--- a/src/locksManager.js
+++ b/src/locksManager.js
@@ -47,14 +47,7 @@ var LocksManager = GObject.registerClass({
     }
 
     isUnlocked(lock) {
-        try {
-            const [variant] = this._proxy.GetSync(lock);
-            return !variant.deep_unpack()
-                           .locked
-                           .deep_unpack();
-        } catch (error) {
-            return false;
-        }
+        return !this._proxy.getDictValueSync(lock, 'locked', true);
     }
 
     setUnlocked(lock) {
--- a/src/toolbox.js
+++ b/src/toolbox.js
@@ -1,11 +1,14 @@
 /* exported Toolbox */
 
 const {GObject, Gtk} = imports.gi;
+const GameState = imports.gameState;
 const Gettext = imports.gettext;
+const Signals = imports.signals;
 
 const _ = Gettext.gettext;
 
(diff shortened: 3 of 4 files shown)


src/toolbox.js
```diff
@@ -61,6 +64,10 @@ var Toolbox = GObject.registerClass({
         this._buttonReset.connect('clicked', this._onResetClicked.bind(this));
         this._buttonMinimize.connect('clicked', this._onMinimize.bind(this));
         this.setBusy(false);
+
+        this._gameState = GameState.getDefault();
+        this._gameState.connect('g-signal', this._onStateChanged.bind(this));
```
ptomato: If you use the (unfortunately JS-only) signal that the proxy object creates, then you can do the state change without another sync call to get the state.

See https://github.com/endlessm/hack-hackdex/blob/master/com.endlessm.Hackdex_chapter_one/controller/meshDev.js#L41 for an example

src/toolbox.js
```diff
@@ -113,4 +120,26 @@ var Toolbox = GObject.registerClass({
...
+        this._buttonReset.visible = visible;
+    }
+
+    _onStateChanged(proxy, sender, signal, params) {
+        if (signal !== 'changed')
+            return;
+        const key = params.get_child_value(0).deep_unpack();
+        if (key !== RESET_BUTTON_KEY)
+            return;
+        this._updateResetButton();
```
ptomato: Or alternatively, get the value from `get_child_value(1)` and update the state based on that. It just seems wasteful to do another sync call if you already have the value here.

ptomato (changes requested):
One comment for saving a sync call

src/toolbox.js
tchx84: I can do the same thing with the current approach :) But since I _have_ to read it while constructing the toolbox, and whole try/unpack boilerplate hurts, is just looked simpler to do it this way. But I can surely save that one call.

src/toolbox.js
tchx84: Yeah, this is what I meant. But, I am curious about how you approach works.

src/locksManager.js
```diff
@@ -47,14 +47,7 @@ var LocksManager = GObject.registerClass({
...
     isUnlocked(lock) {
-        try {
-            const [variant] = this._proxy.GetSync(lock);
-            return !variant.deep_unpack()
-                           .locked
-                           .deep_unpack();
-        } catch (error) {
-            return false;
-        }
+        return !this._proxy.getDictValue(lock, 'locked', true);
```
ptomato: Seems like the code you are replacing had the default value equal to `false`, now it is `true`?

src/gameState.js
```diff
@@ -52,6 +62,9 @@ var getDefault = (function () {
                     });
                 });
             };
+            Object.assign(defaultGameStateProxy, {
```
ptomato: Optional; is there a reason to not define it inline like the other methods here?

ptomato (changes requested):

src/locksManager.js
tchx84: @ptomato yes, and that is correct. !true === false, and that is what should return as default value (note the negation).

src/gameState.js
tchx84: Not really, I can fix it.

PR #82, opened 2018-11-19
T24442

tchx84:
https://phabricator.endlessm.com/T24442


xjuan:
Looks good, the only thing is that the selected cursor does not have the focus when you first open the popup, so we either need to sync the focus with highlighted item or simply make sure items can not have focus

I think I prefer the later

xjuan:
Looks good!

CR Pass

PR #83, opened 2018-11-21
T24415

cosimoc:
https://phabricator.endlessm.com/T24415


tchx84:
One thing we should fix and another to that we should clarify:

(problem) In this implementation, we apply the overlay when the toolbox says something changed, using the model's `changed` property... but when we change something and we click on the flip button to go back, the model takes a "snapshot" of the current values, so these values can be considered "the original values" the next time we flip to modify the app. Because of this, the `changed` property is set to `false`, so the overlay disappears before the flip-back animations starts, which looks like an visual artifact.

(needs for clarification) The reset button (can be a tricky one). According to the specs, the overlay should only be displayed when there are changes, but I think the current implementation is a bit limited in the following scenarios:

a) the app is flipped for the first time, so the model values and the default values are the same, if all I do is pressing the reset button the overlay is not displayed because of course nothing changed, but if I change something the overlay is displayed  and everything is good until now... so then when I press reset again, the values are restored so technically there aren't any changes anymore, but the overlay is still being displayed... Similarly, if I change one value and then restore it manually to the original value, the overlay is still displayed.

b) we are flipping the app after we already applied changes, so the model values and default values do not match anymore, in this case clicking on the button displays the overlay because the app does change from its immediate previous state, but if compared to the previous scenario is kind of confusing?

I think we should partially fallback to the previous behavior where the overlay is always displayed when filliping, so fixes the problem, and at least until we can clarify the behavior of what `changed` should be (or use a different property for implementing this, like `changed-from-original-values`).

tchx84:
@erikos if you agree, we can go with my proposal of having the new visuals for the overlay but keeping the same behavior until we can properly fix the issues I mentioned.

I created a PR for it, which is basically the second commit of this PR with a very small change https://github.com/endlessm/hack-toolbox-app/pull/85

tchx84:
Since @erikos merged my PR, I am closing this for now, so this doesn't get merged accidentally. So, we can discuss with @ptomato  what would be the best way of tracking changes in the framework toolbox model, and we can achieve the behavior design wants (and without visual artifacts).

PR #84, opened 2018-11-21
T24522

tchx84:
https://phabricator.endlessm.com/T24522

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -74,6 +74,10 @@ toolbox revealer {
     padding: 6px;
 }
 
+.reset:hover {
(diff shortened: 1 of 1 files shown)


data/application.css
```diff
@@ -74,6 +74,10 @@ toolbox revealer {
     padding: 6px;
 }
 
+.reset:hover {
+   background-color: rgba(29, 112, 213, 0.795);
```
manuq, line 78: According to my calcs this should be `rgba(26, 100, 191, 0.795)`. But it depends on how do you make the original color lighter. So I'll merge it as is.

manuq (approved):

PR #85, opened 2018-11-22
T24415

tchx84:
https://phabricator.endlessm.com/T24415


tchx84:
This is an alternative PR from https://github.com/endlessm/hack-toolbox-app/pull/83 (these 2 should not be merged together.)

PR #87, opened 2018-11-22
WIP: Lockscreen: use a video for animations

xjuan:
Instead of implementing every animations with css or code, simply use a video
with a key color as transparency.

Videos and stills goes under data/lockscreens/lockId/

https://phabricator.endlessm.com/T24541


tchx84:
I see the building blocks are there, and this is looking surprisingly simple (kudos!).

But, I see some issues that can be solved by clarifying how the user interactions with the lock-screen must work. I will explain below how the interaction works, so my in-line comments will make more sense:

0. (no key available, lock never unlocked before) the user sees the lock-screen, clicking on the lock-screen does nothing (maybe in the future it can play a ACCESS DENIED sound).
1. (key is given, lock never unlocked before) the user _still_sees the lock-screen, but with a different image so the user knows that "now it can be opened". The user clicks the lock-screen, THEN the animation plays and the lock-screen disappears.
3. (key is given, lock already unlocked) the user does not sees the lock-screen nor the animation anymore. When the user opens the toolbox, the contents behind the lock screen are shown right-away.

src/lockscreen.js
```diff
@@ -23,21 +26,42 @@ var Lockscreen = GObject.registerClass({
...
             return Gdk.EVENT_PROPAGATE;
         });
 
+        this._css_provider = new Gtk.CssProvider();
+        this._style = this._overlay.get_style_context();
+        this._style.add_provider(
+            this._css_provider,
+            Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
+        );
+        this._playbin = new Playbin({ visible: true });
```
tchx84: Keep in mind that the lock-screen image or animation, should _only_ be shown if the lock screen has never been unlocked before.

src/lockscreen.js
```diff
@@ -23,21 +26,42 @@ var Lockscreen = GObject.registerClass({
...
+        this._overlay.add(this._playbin);
+
+        this._updateUI();
+
         this._manager = Gio.Application.get_default().locksManager;
         this._keyChangedId = 0;
         this._lockChangedId = 0;
         this.connect('overlay-clicked', this._onClicked.bind(this));
+
+        GLib.timeout_add_seconds(GLib.PRIORITY_LOW, 1, () => {
```
tchx84: Don't forget to remove this hack!

src/lockscreen.js
```diff
@@ -86,12 +110,34 @@ var Lockscreen = GObject.registerClass({
         if (!this._key)
             return;
         if (this._manager.hasKey(this._key)) {
+            this._style.remove_class('locked');
```
tchx84: What needs to happen here is that the lock-screen should update the image to another image that tells the user that "know it can be opened".

src/lockscreen.js
```diff
@@ -86,12 +110,34 @@ var Lockscreen = GObject.registerClass({
...
+    _getAssetsDir () {
+        if ('_lock' in this) {
+            var dir = Gio.File.new_for_path('/app/share/lockscreens/' + this._lock);
+            if (dir.query_exists(null))
+                return this._lock;
         }
+        return 'default';
     }
 
     _updateLockStateWithLock() {
```
tchx84, line 131: I would simply move 

```
        if (!this._lock)
            return;
```

To here so we don't need to repeat it.

src/lockscreen.js
```diff
@@ -86,12 +110,34 @@ var Lockscreen = GObject.registerClass({
...
+        var dir = this._getAssetsDir();
+
+        try {
+            this._css_provider.load_from_data(`
+widget.locked {
+    background-color: red;
+    background: url('file:///app/share/lockscreens/${dir}/locked.png') no-repeat center;
+}`
+                );
+        } catch (e) {
```
tchx84: Just curious, can this really happen?

src/lockscreen.js
```diff
@@ -86,12 +110,34 @@ var Lockscreen = GObject.registerClass({
...
+            logError(e, 'Failed to load css');
+        }
+        this._playbin.uri = `file:///app/share/lockscreens/${dir}/animation.webm`;
+
         if (!this._lock)
             return;
         if (!this._manager.isUnlocked(this._lock))
```
tchx84, line 139: What needs to happen here is that:

a)  if the lock-screen `_lock` says its "locked" but the GSS says is now unlocked, then it means that the lock _just_ got unlock, therefore, the animation should be played.
 or

b) if the lock-screen `_lock` says its already unlock, and the GSS agrees, then it means it was already unlock, therefore we shouldn't even dislay the lock-screen anymore, neither the animation.

src/lockscreen.js
```diff
@@ -116,21 +162,14 @@ var Lockscreen = GObject.registerClass({
...
         } else {
-            style.remove_class('locked');
-
-            // hide widget after transition completes, so that it doesn't
-            // continue to intercept clicks
-            GLib.timeout_add(GLib.PRIORITY_DEFAULT, FADE_OUT_TIME_MS, () => {
-                this.remove(this._overlay);
-                return GLib.SOURCE_REMOVE;
-            });
+            this._playbin.play();
```
tchx84, line 176: Keep in mind that, the animation should only be displayed under the specific condition where 1) the lock-screen was being displayed as locked, 2) the user by clicking or maybe the clubhouse by interacting with the GSS, has update the lock GSS state to be unlocked.

So we shouldn't be playing or stopping the animation here.

src/lockscreen.js
tchx84: Making it visible here is messing with some toolboxes logic where some panels/lock should only shown after another has been unlocked.

src/playbin.js
```diff
@@ -0,0 +1,70 @@
...
+
+        this._started = false;
+
+        this._playbin = Gst.parse_launch('playbin name=playbin video-sink=gtksink');
+        this._playbin.videoFilter = Gst.parse_bin_from_description('videobox name=videobox ! alpha method=green', true);
+        this._videobox = this._playbin.videoFilter.get_by_name('videobox');
+
+        /* TODO: calculate right autocrop values */
+        this._videobox.top = 38;
+        this._videobox.bottom = 38;
```
tchx84: Does this explain (?)

screenshot from 2018-11-23 10-06-29 1

src/lockscreen.js
tchx84: To be more clear, we are missing one asset per lock-screen now.

tchx84:
One last comment, run the lint tool!

tchx84:
Looking great, some comments/questions:

Non-critical:
* Is that (see image below) the intended (for this PR) lockscreen image ?
* Remove prints!
* Is the fizzics.2 lock assets missing?
* We need to add symlinks for encyclopedia.en as well (rebase on master, and you will note that we support it now).
* See Anitas email regarding moving some locks assets around.

Critical:
* There is a strange regression with the lvl2 lock from the framework toolbox: sometimes the animation won't play, and then (more importantly) the lockscreen seems to disappear visually, but the contents can't be accessed, e.g you can't click/select the panel options. To reproduce it start the dinosaurs toolbox with the following state:

```
{
  "item.key.framework.1" : {
    "consume_after_use" : true,
    "used" : false
  },
  "item.key.framework.2" : {
    "consume_after_use" : true,
    "used" : false
  },
  "item.key.framework.3" : {
    "consume_after_use" : true,
    "used" : false
  }
}
```
 
screenshot from 2018-11-30 08-33-38 1

src/com.endlessm.HackToolbox.src.gresource.xml
```diff
@@ -33,6 +33,7 @@
     <file>OperatingSystemApp/oscursormodel.js</file>
     <file>OperatingSystemApp/oswobblymodel.js</file>
     <file>OperatingSystemApp/toolbox.js</file>
+    <file>playbin.js</file>
```
cosimoc: Alphabetical order

src/playbin.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    add_class (klass) {
+        this._style.add_class(klass);
+    }
+
+    remove_class (klass) {
+        this._style.remove_class(klass);
+    }
```
cosimoc: I don't find it worth it to have these special methods here, since the playbin is a widget itself. Clients can simply call `get_style_context()` on it directly as they would on other widgets.

src/playbin.js
```diff
@@ -0,0 +1,157 @@
...
+    Signals: {
+        clicked: {},
+        start: {},
+        done: {},
+    },
+}, class Playbin extends Gtk.EventBox {
+    _init(props) {
+        super._init(props);
+
+        this._style = this.get_style_context();
```
cosimoc: There's no need to cache this as a private member in JS, since GTK already does that for us.

src/playbin.js
```diff
@@ -0,0 +1,157 @@
...
+
+    _updateCropArea () {
+        const width = this._video_widget.get_allocated_width();
+        const height = this._video_widget.get_allocated_height();
+
+        if (width < 2 || height < 2 || !this._video_width || !this._video_height)
+            return;
+
+        const crop_x = (this._video_width - width) / 2;
+        const crop_y = (this._video_height - height) / 2;
```
cosimoc, line 68: Is it worth it here to check whether the crop_x/crop_y values are already the same set on the videocrop and bail our early, or does videocrop already optimize that?

src/playbin.js
```diff
@@ -0,0 +1,157 @@
...
+            const [, newstate] = msg.parse_state_changed();
+
+            if (!this._started && newstate === Gst.State.PLAYING) {
+                this._video_widget.show_now();
+                this._started = true;
+                this.emit('start');
+            }
+        }
+
+        return GLib.SOURCE_CONTINUE;
```
cosimoc, line 130: Nitpick: `GstBusFunc` is defined to return `true` or `false`

src/playbin.js
```diff
@@ -0,0 +1,157 @@
...
+    play () {
+        this._ensurePlaybin();
+
+        this._started = false;
+        this._playbin.uri = this.uri;
+        this._playbin.set_state(Gst.State.PLAYING);
+    }
+
+    stop () {
+        this._playbin.set_state(Gst.State.NULL);
```
cosimoc: Probably worth it here to check if the playbin exists.

data/extra-install.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+FRAMEWORKAPPS="
+	com.endlessm.dinosaurs.en
+	com.endlessm.encyclopedia.en
+"
+
+LOCKSCREENSDIR=/app/share/lockscreens
```
cosimoc: It would be cool to respect the build prefix here.

Also I am not sure that this script is necessary to be honest, or the best way to solve this problem. The toolbox should already know that a framework app uses lockscreens called "framework", and I'd rather have the toolbox look for that asset directly rather than adding all these symlinks here.

@ptomato perhaps you can help suggesting how to do that.

src/lockscreen.js
```diff
@@ -1,51 +1,60 @@
...
+        this._updateUI();
     }
 
     get locked() {
         return this._locked;
     }
 
     set locked(value) {
-        if ('_locked' in this && this._locked === value)
+        if (this._locked === value)
```
cosimoc: I'd like to preserve the original `if('$prop' in this)` pattern here, unless there's a reason not to

src/lockscreen.js
```diff
@@ -72,34 +81,49 @@ var Lockscreen = GObject.registerClass({
...
+        if (this._lock === lock)
             return;
         if (this._lockChangedId !== 0)
             this._manager.disconnect(this._lockChangedId);
         this._lockChangedId = this._manager.connect(
             `changed::${lock}`, this._updateLockStateWithLock.bind(this));
         this._lock = lock;
+
+        var path = `/app/share/lockscreens/${lock}`;
+        var dir = Gio.File.new_for_path(path);
```
cosimoc: I think these could be `const`?

src/lockscreen.js
```diff
@@ -72,34 +81,49 @@ var Lockscreen = GObject.registerClass({
...
-        if ('_lock' in this && this._lock === lock)
+        if (this._lock === lock)
             return;
         if (this._lockChangedId !== 0)
             this._manager.disconnect(this._lockChangedId);
         this._lockChangedId = this._manager.connect(
             `changed::${lock}`, this._updateLockStateWithLock.bind(this));
         this._lock = lock;
+
+        var path = `/app/share/lockscreens/${lock}`;
```
cosimoc: It would be great to respect the build prefix, or do some kind of relative lookup.

src/lockscreen.js
```diff
@@ -1,51 +1,60 @@
...
+        this._lock = null;
+
         super._init(props);
 
-        this._overlay = new Gtk.EventBox({
+        this._playbin = new Playbin({
             expand: true,
-            visible: true,
+            noShowAll: true,
+            visible: false,
```
cosimoc: This is the default

src/lockscreen.js
```diff
@@ -72,34 +81,49 @@ var Lockscreen = GObject.registerClass({
...
 
     _updateLockStateWithKey() {
         if (!this._key)
             return;
-        if (this._manager.hasKey(this._key)) {
-
-            /* TODO update the lock-screen graphics with can-be-opened image */
-        }
+        if (this._assetsPath && this._manager.hasKey(this._key))
+            this._playbin.setBackground(`file://${this._assetsPath}/has-key`);
```
cosimoc: Should we also set the `no-key` state here in the `else` case?

src/lockscreen.js
```diff
@@ -72,34 +81,49 @@ var Lockscreen = GObject.registerClass({
...
+        if (!this._lock) {
+            this._playbin.add_class('no-lock');
             return;
+        }
+
+        this._playbin.remove_class('no-lock');
+        if (this._assetsPath) {
+            var bg = this._manager.hasKey(this._key) ? 'has-key' : 'no-key';
+            this._playbin.setBackground(`file://${this._assetsPath}/${bg}`);
+        }
```
cosimoc: Perhaps we can call `_updateLockStateWithKey()` here? It's unclear to me if we need to do it here in addition to what we do in `_updateLockStateWithKey()`.

cosimoc: I'd also try to centralize every UI state change in `_updateUI()`, it's difficult to follow all the flow otherwise.

cosimoc (changes requested):

src/lockscreen.js
ptomato: `eslint --fix` should fix that

data/extra-install.sh
ptomato: To respect the prefix in this script you can look in the environment variable `MESON_INSTALL_PREFIX`, it will be set when running in the script.

I think it would be difficult to special-case the lockscreen names. It's already determined that the locks are named per-app, since the key has to open each app's lock separately. You could name the assets according to the key IDs rather than the lock IDs, but then you would not be able to have a separate asset for the hackdex lock (because that is also opened by the framework key.) Basically I think there will have to be annoying architecture somewhere, and it might as well be here, so that we can keep the lockscreen asset names equal to the lock IDs.

src/playbin.js
cosimoc: It already optimizes that.

PR #89, opened 2018-11-27
T24615

tchx84:
https://phabricator.endlessm.com/T24615


src/Fizzics/level2.js
```diff
@@ -230,10 +247,19 @@ var FizzicsLevel2 = GObject.registerClass({
             const modelProp = props[prop];
             const value = this._getValueForScope(prop, modelProp);
             const type = typeof value;
+            const options = this._getOptionsForScopeValue(scope, prop);
+            const base_message = `Unknown value ${scope[prop]} for ${prop}`;
```
cosimoc: Please use camel case

src/Fizzics/level2.js
```diff
@@ -230,10 +247,19 @@ var FizzicsLevel2 = GObject.registerClass({
...
+            print(prop, props[prop], options);
```
cosimoc: I believe this is a debug leftover?

cosimoc (changes requested):

src/Fizzics/level2.js
tchx84: Yes, good catch.

PR #94, opened 2018-11-29
Fix card hover state

masimka:
Add new rules into `gen.js` for the hover state.

https://phabricator.endlessm.com/T24673


ptomato:
Thanks.

PR #100, opened 2018-12-03
Fix homepage characters cards

xrgranados:
https://phabricator.endlessm.com/T24721


ptomato:
This is not working correctly on 3200x1800@2x, in fact it looks worse than before on that resolution. I think it's best that we just leave this as-is until we can refactor the cards so that they don't use an asset sprite.

ptomato (changes requested):
I think we can close this, as we discussed that it should be solved by not using an asset sprite.

PR #101, opened 2018-12-04
Playbin: fix playback start glitch

xjuan:
https://phabricator.endlessm.com/T24734

### DIFF
--- a/src/playbin.js
+++ b/src/playbin.js
@@ -1,22 +1,14 @@
 /* exported Playbin */
 
-const {Gdk, GLib, GObject, Gtk, Gst} = imports.gi;
+const {GLib, GObject, Gtk, Gst} = imports.gi;
 
-var Playbin = GObject.registerClass({
-    CssName: 'playbin',
-    Properties: {
-        uri: GObject.ParamSpec.string('uri', 'URI', '',
-            GObject.ParamFlags.READWRITE,
-            ''
-        ),
-    },
-    Signals: {
-        clicked: {},
-        start: {},
-        done: {},
-    },
-}, class Playbin extends Gtk.EventBox {
+const Lock = GObject.registerClass({
+    CssName: 'lock',
+}, class Lock extends Gtk.Button {
     _init(props) {
+        props.canDefault = false;
+        props.canFocus = false;
+
         super._init(props);
 
         this._css_provider = new Gtk.CssProvider();
@@ -24,10 +16,48 @@ var Playbin = GObject.registerClass({
             this._css_provider,
             Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
         );
+    }
 
-        this.connect('button-release-event', () => {
+    set background (url) {
+        try {
+            this._css_provider.load_from_data(`
+lock {
+    background: url('${url}') no-repeat center;
+    transition: background 1s ease;
+}`
+            );
+        } catch (e) {
+            logError(e, 'Failed to load css');
+        }
+    }
+});
+
+var Playbin = GObject.registerClass({
+    CssName: 'playbin',
+    Signals: {
+        clicked: {},
+        done: {},
+    },
+}, class Playbin extends Gtk.Stack {
+    _init(props) {
+        this._uri = null;
+
+        props.transitionType = props.transitionType || Gtk.StackTransitionType.CROSSFADE;
+        props.transitionDuration = props.transitionDuration || 100;
+
+        super._init(props);
+
+        /* Use a button to catch the click event */
+        this._button = new Lock({visible: true});
+        this.add(this._button);
+
+        this._button.connect('clicked', () => {
             this.emit('clicked');
-            return Gdk.EVENT_PROPAGATE;
+        });
+
+        /* Update crop values on size allocation */
+        this.connect('size-allocate', (self, alloc) => {
+            this._updateCropArea(alloc);
         });
     }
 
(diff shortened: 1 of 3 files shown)


src/playbin.js
```diff
@@ -25,9 +26,22 @@ var Playbin = GObject.registerClass({
...
+        });
+
+        /* Make button invisible */
+        this._button.connect('draw', () => {
+            return true;
+        });
+
+        /* Update crop values on size allocation */
+        this.connect('size-allocate', (self, alloc) => {
+            self._updateCropArea(alloc);
```
cosimoc: You can directly call `this._updateCropArea()`

src/playbin.js
```diff
@@ -44,41 +58,33 @@ var Playbin = GObject.registerClass({
...
-
-        if (!this._started && this._state === Gst.State.PLAYING) {
-            this._started = true;
-            this.emit('start');
-        }
     }
 
-    _updateSizeFromStreamCollection (collection) {
+    _onStreamsSelected (msg) {
+        var collection = msg.parse_streams_selected();
```
cosimoc: Should this not be `const`?

src/playbin.js
```diff
@@ -102,29 +108,43 @@ var Playbin = GObject.registerClass({
...
+
+    _onStateChanged (msg) {
+        if (this._started)
+            return;
+
+        const [, newstate] = msg.parse_state_changed();
+
+        if (newstate === Gst.State.PLAYING) {
+            this.set_visible_child(this._video_widget);
+            this.get_style_context().remove_class('locked');
```
cosimoc: I'd rather you set `locked = false` directly here, so there are not two places where style classes are tweaked... But also I would consider emitting a signal from here and setting it from the lockscreen through `_updateUI()` instead.

src/playbin.js
```diff
@@ -142,22 +162,34 @@ playbin.locked {
...
+        this._playbin.uri = this._uri;
         this._playbin.set_state(Gst.State.PLAYING);
     }
 
     set locked(value) {
         const style = this.get_style_context();
         if (value)
             style.add_class('locked');
-        else
+        else if (!this.uri)
```
cosimoc: Why this check here? I'd rather move this check to whichever caller needed it, and keep this as a simple function.

cosimoc (changes requested):

src/playbin.js
```diff
@@ -1,22 +1,23 @@
...
-const {Gdk, GLib, GObject, Gtk, Gst} = imports.gi;
+const {GLib, GObject, Gtk, Gst} = imports.gi;
 
 var Playbin = GObject.registerClass({
     CssName: 'playbin',
     Properties: {
         uri: GObject.ParamSpec.string('uri', 'URI', '',
             GObject.ParamFlags.READWRITE,
-            ''
+            null
```
ptomato: JS doesn't have the concept of NULL strings, I'm not sure what will happen if you do this, but I think it might get marshalled as the empty string anyway, when setting the default value of the property. But I am pretty sure it's fine just to set `this._uri = null` in the `_init()` method as you do below, and leave this as-is.

src/playbin.js
```diff
@@ -25,9 +26,22 @@ var Playbin = GObject.registerClass({
...
+        this._button = new Gtk.Button({visible: true});
+        this.add(this._button);
+
+        this._button.connect('clicked', () => {
             this.emit('clicked');
-            return Gdk.EVENT_PROPAGATE;
+        });
+
+        /* Make button invisible */
+        this._button.connect('draw', () => {
```
ptomato: `() => true` is shorter here

src/playbin.js
ptomato: I wonder why the linter in travis CI didn't catch this...

ptomato (commented):
Couple of drive-by comments

src/playbin.js
xjuan: yeah I can, but then the linter will complain I did not used self right?

cosimoc: I don't think so, but if it does, you can simply add `void self;` or something like that

ptomato: if you use `alloc` then it will not complain about `self`

src/playbin.js
```diff
@@ -1,33 +1,71 @@
...
+const {GLib, GObject, Gtk, Gst} = imports.gi;
+
+const Lock = GObject.registerClass({
+    CssName: 'lock',
+}, class Lock extends Gtk.Button {
+    _init(props) {
+        super._init(props);
+
+        this.canDefault = false;
+        this.canFocus = false;
```
cosimoc: It would be nicer to fold these in the construction `props` dictionary above

src/playbin.js
```diff
@@ -1,33 +1,71 @@
...
+
+        /* Make button invisible */
+        this.connect('draw', cr => {
+            const style = this.get_style_context();
+            const c = this.get_clip();
+            Gtk.render_background(style, cr, c.x, c.y, c.width, c.height);
```
cosimoc: How does this make the button invisible like the comment says? Also, I understand that code is here because the theme sets a background on the button, but buttons always draw a background in GTK, so you should not need the override...

src/playbin.js
```diff
@@ -1,33 +1,71 @@
...
+
         super._init(props);
 
-        this._css_provider = new Gtk.CssProvider();
-        this.get_style_context().add_provider(
-            this._css_provider,
-            Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
-        );
+        this.transitionType = Gtk.StackTransitionType.CROSSFADE;
+        this.transitionDuration = 100;
```
cosimoc: Please fold these into the `props` construction property dict.

src/playbin.js
cosimoc: Cool! You should be able to change this then

src/playbin.js
```diff
@@ -102,63 +132,81 @@ var Playbin = GObject.registerClass({
...
     set locked(value) {
-        const style = this.get_style_context();
-        if (value)
-            style.add_class('locked');
-        else
-            style.remove_class('locked');
+        if (value) {
+            this._button.show();
+            this.set_visible_child(this._button);
+        } else if (!this._uri) {
```
cosimoc: I am not sure why this check needs to be here... Feels to me like it'd belong in whoever calls `locked = false`. I would like to keep this function "dumb" and just set the locked value it's passed, without state checks.

cosimoc (changes requested):

PR #103, opened 2018-12-06
T24801

tchx84:
https://phabricator.endlessm.com/T24801

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -206,8 +206,10 @@ var Lockscreen = GObject.registerClass({
         }
         if (!this._key || !this._lock)
             return;
-        if (!this._manager.hasKey(this._key))
+        if (!this._manager.hasKey(this._key)) {
+            SoundServer.getDefault().play('hack-toolbox/lockscreen/inactive');
             return;
+        }
 
         /* We are going to need to playback an animation */
         if (this._openURI)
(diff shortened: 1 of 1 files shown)


src/lockscreen.js
```diff
@@ -130,6 +142,27 @@ var Lockscreen = GObject.registerClass({
...
     }
 
+    _playActiveSound() {
+        if (!this._locked)
+            return;
+        if (!this._playbin.hasKey)
+            return;
+        if (this._activeSoundID === 'pending')
+            return;
+        this._activeSoundID = 'pending';
```
ptomato: I think for this to work correctly you have to do the other half of it as well, like we do in the shell; check for `pending` in `_stopActiveSound` and set it to `cancelled`, and check for `cancelled` here as well.

ptomato (changes requested):

src/lockscreen.js
tchx84: I am actually trying something completely different in parallel now, since this UX has one problem I discussed here https://phabricator.endlessm.com/T24801#658274, and yeah, what I am trying now kind of uses that, but I am moving it to the sound wrapper.

tchx84:
Polished this a bit more, so make sure to update locally.

PR #106, opened 2018-12-10
T24852

tchx84:
https://phabricator.endlessm.com/T24852


ptomato:
This is a reasonable solution but given that we've had a request to control this behaviour in the sound server's metadata file, I think it would be better to implement it that way. See comment in Phabricator.

xjuan:
Outdated

PR #111, opened 2018-12-14
lockscreen: only play one active sound at the same time

cfoch:
A general rule:

* The active loop sound must be played IFF there is
  at least one active lockscreen.

To achieve this:

* Use the new SoundServer.playUnique API to ensure that,
  given many active lockscreens, only 1 will be played.
  This is done by tagging the "hack-toolbox/lockscreen/active"
  sound with "overlap-behavior" as "ignore" in the
  hack-sound-server metadata file.

* In order to avoid a complex inter-lockscreen coordination
  scheme, rely on the simple rule that, whenever a lockscreen
  is unlocked, it is an opportunity for another lockscreen to
  play its own active sound.

https://phabricator.endlessm.com/T24852

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -93,7 +93,6 @@ var Lockscreen = GObject.registerClass({
         this._keyChangedId = this._manager.connect(
             `changed::${key}`, () => {
                 this._updateLockStateWithKey();
-                this._playActiveSound();
             });
         this._key = key;
         this._updateLockStateWithKey();
@@ -108,7 +107,10 @@ var Lockscreen = GObject.registerClass({
             return;
         this._untrackLockChanges();
         this._lockChangedId = this._manager.connect(
-            `changed::${lock}`, this._updateLockStateWithLock.bind(this));
+            `changed::${lock}`, () => {
+                this._updateLockStateWithLock();
+                this._stopActiveSound();
+            });
         this._lock = lock;
         this._updateLockStateWithLock();
     }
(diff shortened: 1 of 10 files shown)


src/lockscreen.js
```diff
@@ -41,9 +41,8 @@ var Lockscreen = GObject.registerClass({
         this._activeSoundID = null;
         this._updateUI();
 
-        this._playbin.connect('realize', () => {
-            this._playActiveSound();
-        });
+        this._manager.connect_after('changed', this._playActiveSound.bind(this));
+        this._playbin.connect('realize', this._playActiveSound.bind(this));
```
xjuan: Why do we need to start the sound on both these places?

Also there seems to be a regression where the sound does not stop after the animation starts (I can reporduce it on the system app)

xjuan (changes requested):

src/lockscreen.js
cfoch: > Why do we need to start the sound on both these places?

This is what Martin added. I am not totally sure TBH if `        this._manager.connect_after('changed', this._playActiveSound.bind(this));` should be considered.

> Also there seems to be a regression where the sound does not stop after the animation starts (I can reporduce it on the system app)

Are you testing it against https://github.com/endlessm/hack-sound-server/pull/55 ?

erikos: @xjuan can you take another look with the #55 sound server?

xjuan: Yes I did,, see comment in phabricator

src/lockscreen.js
```diff
@@ -4,12 +4,17 @@
...
 const {Playbin} = imports.playbin;
+const {Toolbox} = imports.toolbox;
 const SoundServer = imports.soundServer;
 
 var Lockscreen = GObject.registerClass({
     GTypeName: 'Lockscreen',
     CssName: 'lockscreen',
     Properties: {
+        toolbox: GObject.ParamSpec.object(
+            'toolbox', 'Toolbox', '',
```
tchx84, line 15: is accessing the `target-app-id` from the lockscreen the goal here? If so, we should just add a `target-app-id` property and avoid all that unnecessary toolbox refactoring.

Note that:
1.  the window already carries that property https://github.com/endlessm/hack-toolbox-app/blob/master/src/window.js#L8
2.  there is already logic in place to pass that property to the lockscreen, e.g. https://github.com/endlessm/hack-toolbox-app/blob/master/src/framework/controlPanel.js#L46 (same thing happens for all toolboxes).

src/lockscreen.js
```diff
@@ -174,38 +176,21 @@ var Lockscreen = GObject.registerClass({
...
-            return;
-        if (this._activeSoundID === 'cancel')
-            return;
-        if (this._activeSoundID === 'pending') {
-            this._activeSoundID = 'cancel';
-            return;
+        if (this._activeSoundID !== null) {
+            SoundServer.getDefault().stopAtBusName(
+                this._activeSoundID,
+                this.toolbox.target_app_id
```
tchx84, line 190: Here (and similar places) it would only need to access it own `target-app-id` property.

tchx84 (changes requested):
Left a couple questions/comments.

src/lockscreen.js
cfoch: Thanks. What you propose here is better.

tchx84:
If I understood correctly this PR is no longer needed.

PR #112, opened 2018-12-14
playbin: use newly added ready-to-show signal

xjuan:
Use new GtkGstBaseWidget::ready-to-show signal to know exactly when to show
the video widget without rendering any blank frames.

Upstream https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/issues/528

https://phabricator.endlessm.com/T24812

### DIFF
--- a/src/playbin.js
+++ b/src/playbin.js
@@ -87,6 +87,15 @@ var Playbin = GObject.registerClass({
         this._video_widget.show();
         this.add(this._video_widget);
 
+        this._has_ready_to_show = !!GObject.signal_lookup('ready-to-show',
+            this._video_widget.constructor.$gtype);
+
+        if (this._has_ready_to_show) {
+            this._video_widget.connect('ready-to-show', () => {
+                this.set_visible_child(this._video_widget);
+            });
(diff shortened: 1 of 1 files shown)


src/playbin.js
```diff
@@ -87,6 +87,15 @@ var Playbin = GObject.registerClass({
         this._video_widget.show();
         this.add(this._video_widget);
 
+        this._has_ready_to_show = !!GObject.signal_lookup('ready-to-show',
+            this._video_widget.constructor.$gtype);
```
cfoch, line 91: Does  "!!" do something special? 
Isn't this equivalent to
```
        this._has_ready_to_show = GObject.signal_lookup('ready-to-show',
            this._video_widget.constructor.$gtype);
```
?

ptomato: `!!` converts to a boolean. It's almost equivalent, but having a boolean is slightly easier when debugging because you don't have to wonder what `this._has_ready_to_show === 379` means.

cfoch: Then it's okay as it is.

PR #113, opened 2018-12-14
Metrics in code panel

ptomato:
https://phabricator.endlessm.com/T24429

### DIFF
--- a/src/codeview.js
+++ b/src/codeview.js
@@ -1,8 +1,12 @@
 /* exported Codeview */
 
-const {Gdk, GLib, GObject, Gtk, GtkSource} = imports.gi;
+const {EosMetrics, Gdk, Gio, GLib, GObject, Gtk, GtkSource, HackToolbox} = imports.gi;
+const ByteArray = imports.byteArray;
 
 const SoundServer = imports.soundServer;
+void imports.utils;  // pull in promisified functions
+
+const CODEVIEW_ERROR_EVENT = 'e98aa2b8-3f11-4a25-b8e9-b10a635df121';
 
 // Can add more, e.g. WARNING, SUGGESTION
 const MarkType = {
@@ -174,6 +178,8 @@ var Codeview = GObject.registerClass({
         this._compileTimeout = null;
         this._numErrors = 0;
         this._ambientMusicID = null;
+        this._inMetricsEvent = false;
+        this._lastCompiledText = null;
     }
 
     get text() {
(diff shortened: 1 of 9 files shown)


src/codeview.js
```diff
@@ -329,6 +335,58 @@ var Codeview = GObject.registerClass({
         return GLib.SOURCE_REMOVE;
     }
 
+    // returns a diff in the form of an ed-script, which is the format output by
+    // the diff tool that takes the fewest bytes. It is forward-only; you cannot
+    // apply it in reverse.
+    async _diffFromLastCompiledText() {
```
xjuan, line 341: any particular reason why not send the whole source code compressed?

xjuan (commented):
Last but not least, is not this a privacy concern to send private text to the metrics sever?

src/codeview.js
```diff
@@ -329,6 +335,58 @@ var Codeview = GObject.registerClass({
...
+        const eventKey = new GLib.Variant('(ss)',
+            [Gio.Application.get_default().applicationId, userFunction]);
+
+        if (this._numErrors && !this._inMetricsEvent) {
+            const payload = new GLib.Variant('(sssa(suquq))', [
+                Gio.Application.get_default().applicationId,
+                userFunction,
+                this._buffer.text,
+                results.map(({start, end, message}) => [message, start.line,
+                    start.column, end.line, end.column]),
```
xjuan: end could be undefined like in Fizzics
```
(com.endlessm.HackToolbox:3): Gjs-WARNING **: 13:47:28.237: JS ERROR: Error while recording error metrics: TypeError: end is undefined
_recordErrorMetrics/payload<@resource:///com/endlessm/HackToolbox/js/codeview.js:371:35
_recordErrorMetrics@resource:///com/endlessm/HackToolbox/js/codeview.js:370:17
setCompileResults@resource:///com/endlessm/HackToolbox/js/codeview.js:423:9
_compile@resource:///com/endlessm/HackToolbox/js/Fizzics/level2.js:194:13
_init/<@resource:///com/endlessm/HackToolbox/js/Fizzics/level2.js:30:13
compile@resource:///com/endlessm/HackToolbox/js/codeview.js:333:9
main@resource:///com/endlessm/HackToolbox/js/main.js:30:12
run@resource:///org/gnome/gjs/modules/package.js:225:12
@/app/bin/com.endlessm.HackToolbox:14:1
```

so we could do something like this
```
                results.map(({start, end, message}) => {
                    end = end || start;
                    return [message, start.line, start.column, end.line, end.column];
                }),
```
or make the last two parameters optionals

xjuan (changes requested):

src/codeview.js
ptomato: It's already compressed when being sent to the server. The reason for storing only a diff of each step is because I thought that these metrics would otherwise fill up the bucket quickly, as they take up more space on disk than most other events.

ptomato:
I don't think it's a privacy concern, as there's nothing private in any of the code. It will not be much different from the pre-filled code that we already put in. Once it becomes possible to write one's own projects and save them, then we should probably stop recording this metric for those more substantial projects.

src/codeview.js
xjuan: Right, lets assume it also saves them compressed :)

xjuan:
Ok, seems to work, the only thing I could not test was the actual diff.
It seems to me that on the first error it should send the whole text and subsequent error only the diff, but that is not the case, it always sent the whole code.

src/codeview.js
ptomato: It doesn't ... they are stored in GVariant byte format

xjuan:
CR Pass

PR #115, opened 2018-12-17
clippyWrapper: Watch reset property to get reset button working.

cfoch:
https://phabricator.endlessm.com/T24930

### DIFF
--- a/src/Fizzics/toolbox.js
+++ b/src/Fizzics/toolbox.js
@@ -27,6 +27,7 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox 
                 new GLib.Variant('(sava{sv})', ['reset', [], {}]),
                 null, Gio.DBusCallFlags.NONE, -1, null
             );
+            this._controlPanel.reset();
         });
     }
 
--- a/src/Fizzics/controlpanel.js
+++ b/src/Fizzics/controlpanel.js
@@ -30,6 +30,10 @@ var FizzicsControlPanel = GObject.registerClass({
             GObject.BindingFlags.SYNC_CREATE | GObject.BindingFlags.INVERT_BOOLEAN);
     }
 
+    reset() {
+        this._level2.reset();
+    }
+
     bindModel(model) {
         this._level1.bindModel(model);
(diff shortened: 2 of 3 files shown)


cfoch:
It works, but I am not sure if this is mergeable. I need feedback.

cosimoc:
I'd like to understand why the roundtrip is needed.
Is it because otherwise you will receive no notification, and so `_onNotify()` here https://github.com/endlessm/hack-toolbox-app/blob/5818894715484e72e41e41fb01a6b78d14d0cef4/src/Fizzics/level2.js#L317 never gets called, or because `_onNotify()` somehow does not regenerate the code correctly? We should be able to invalidate the current codeview state entirely from within the toolbox, since that's what activates the reset action in the first place.

cfoch:
Do not merge it yet.

cfoch:
Yes, this is ready to merge.

src/Fizzics/toolbox.js
```diff
@@ -27,6 +27,7 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
                 new GLib.Variant('(sava{sv})', ['reset', [], {}]),
                 null, Gio.DBusCallFlags.NONE, -1, null
             );
+            this._controlPanel._level2._onNotify();
```
cosimoc: This is a bit punching a hole through the different layers...
The `FizzicsLevel2` class already has access to the model `reset` signal through the `bindModel()` method. I think that would be a much better place to connect to `reset` and call `_onNotify()` from.

cosimoc (changes requested):

src/Fizzics/toolbox.js
cfoch: But the 'reset' signal is emitted by the Toolbox class. Not from the model.

cfoch: And I have no access to the Toolbox from FizzicsLevel2.

cosimoc:
One way to solve that could to add a reset method to the model and move there the code that is responsible for calling the action... or you could find another way to propagate the reset call down to the levels, either by giving them access to the toolbox or through calling a method directly.

PR #117, opened 2018-12-20
model: Fizzics: Add new death sound names

cfoch:
https://phabricator.endlessm.com/T24884

### DIFF
--- a/src/Fizzics/model.js
+++ b/src/Fizzics/model.js
@@ -39,6 +39,9 @@ var SFXS = [
     'horn',
     'drum',
     'beam',
+    'goal',
+    'gem',
+    'win',
 ];
 
 const _indexes = new Array(SPECIES).fill(0)


cfoch:
This is important. Another reason to accept is because otherwise updating values in the codeview will not update widgets values.

src/Fizzics/model.js
```diff
@@ -39,6 +39,9 @@ var SFXS = [
     'horn',
     'drum',
     'beam',
+    'goal',
+    'gem',
+    'win',
```
llopis, line 44: I thought that the 6th sound was the one for the collision of the player ball and the spiky one, not the overall win (that shouldn't be a collision sound).

cosimoc: I think this is just the name... This PR is required for the ticket so I'll merge it not to leave things in an unfinished state. However, please propose another PR in case we need to adjust it.

PR #119, opened 2018-12-28
window: set toplevel window as modal

cosimoc:
This ensures that every dialog, such as GtkColorChooserDialogs or
others, also end up being set as modal, since GTK will read whether
the toplevel is modal to determine whether to make those modal too.

https://phabricator.endlessm.com/T24879


cosimoc:
This does not actually work, since the toolbox has multiple windows and we can't make each of them modal.

PR #122, opened 2019-01-09
Fizzics: Changing friction range to match latest game logic

llopis:
https://phabricator.endlessm.com/T24802

### DIFF
--- a/data/Fizzics/objectEditor.ui
+++ b/data/Fizzics/objectEditor.ui
@@ -9,9 +9,9 @@
     <property name="page_increment">10</property>
   </object>
   <object class="GtkAdjustment" id="adjustmentFriction">
-    <property name="upper">18</property>
-    <property name="value">9</property>
-    <property name="step_increment">1</property>
+    <property name="upper">100</property>
+    <property name="value">5</property>
+    <property name="step_increment">5</property>
     <property name="page_increment">10</property>
   </object>
   <object class="GtkAdjustment" id="adjustmentGravity">
--- a/src/Fizzics/model.js
+++ b/src/Fizzics/model.js
@@ -60,7 +60,7 @@ function _addPropsForIndex(props, index) {
         _propFlags, 0.0, 0.2, 0.1);
     props[`friction-${index}`] = GObject.ParamSpec.double(
         `friction-${index}`, `friction-${index}`, '',
-        _propFlags, 0.0, 18.0, 9.0);
+        _propFlags, 0.0, 100.0, 5.0);
     props[`usePhysics-${index}`] = GObject.ParamSpec.boolean(
         `usePhysics-${index}`, `usePhysics-${index}`, '',
         _propFlags, true);


tchx84 (changes requested):
As I mentioned on slack, you also need to update it on the objectEdtor.ui file.

llopis:
Which part needs to be updated? All I see related to friction is 
        model.bind_property(map['friction'], this._adjustmentFriction, 'value', flags);

src/Fizzics/model.js
```diff
@@ -60,7 +60,7 @@ function _addPropsForIndex(props, index) {
         _propFlags, 0.0, 0.2, 0.1);
     props[`friction-${index}`] = GObject.ParamSpec.double(
         `friction-${index}`, `friction-${index}`, '',
-        _propFlags, 0.0, 18.0, 9.0);
+        _propFlags, 0.0, 100.0, 5.0);
```
tchx84, line 63: Since you haven't (yet) changed the max/min definitions on the client side, I just want to double check that your intention is to set:

max friction: 100.0
min friction: 0.0
default friction: 5.0

This is what both the model.js and .ui are doing.

llopis: That's correct.

data/Fizzics/objectEditor.ui
```diff
@@ -9,8 +9,8 @@
     <property name="page_increment">10</property>
   </object>
   <object class="GtkAdjustment" id="adjustmentFriction">
-    <property name="upper">18</property>
-    <property name="value">9</property>
+    <property name="upper">100</property>
+    <property name="value">5</property>
     <property name="step_increment">1</property>
```
tchx84: Bah, only noticed this when tested today:

We need to increase the `step_increment` from 1 to 10 here and change the `value` from 5 to 10 instead. Two reasons 1) now that the value goes to 100, moving it 1 by 1 takes forever and  2) that is exactly what we already do with a similar one (radius).

tchx84 (changes requested):
One last comment.

data/Fizzics/objectEditor.ui
llopis: Being able to easily set numbers in the 0-10 range is important because how different it is. I'd be afraid that making 5 the default increment would make it harder to fine tune them. Besides, they can type any number if they want to get very larger, right?

tchx84: The range of values goes from 0 to 100, the UI should allow them to move quickly between the whole range. If they need finer tuning, like you just said, the user can type any number between that range.

EDIT: please double check the changes I suggested above.

llopis: I disagree with this, but I'm making the change anyway because I think it's more important to get this PR in than to wait for another round of discussions.

tchx84: To clarify a bit more, if the 0-10 range is more influential than 11-100, then I think that is the problem we should be solving. As it makes less sense to expose such complex function with a simplistic UI tool. It would be better to just expose a range that makes more sense instead.

Noticed you increase step to 5 (instead of 10), but yeah lets merge it as-in now, as we spent enough time here already.

PR #125, opened 2019-01-18
T25025

cfoch:
https://phabricator.endlessm.com/T25025

### DIFF
--- a/src/codeview.js
+++ b/src/codeview.js
@@ -38,11 +38,12 @@ const Buffer = GObject.registerClass(class Buffer extends GtkSource.Buffer {
     _init(props = {}) {
         super._init(props);
         this._lastSelectionLength = 0;
+        this._lastKeypressTimestamp = null;
 
         this._insertHandler = this.connect('insert-text',
-            this.constructor._onInsertText);
+            this._onInsertText.bind(this));
         this._deleteHandler = this.connect('delete-range',
-            this.constructor._onDeleteRange);
+            this._onDeleteRange.bind(this));
         this.connect('mark-set', this._onMarkSet.bind(this));
     }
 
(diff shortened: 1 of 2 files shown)


ptomato (approved):
This code looks good, I will post some questions for David in the Phabricator task but if we decide not to make any changes as a result of the questions, then we can merge this.

src/codeview.js
```diff
@@ -71,24 +72,42 @@ const Buffer = GObject.registerClass(class Buffer extends GtkSource.Buffer {
...
-            sound.play('codeview/keypress/other');
+            sound.playFull('codeview/keypress/other', options);
     }
 
-    static _onDeleteRange(buffer, start, end) {
+    _onDeleteRange(buffer, start, end) {
         const deletedLength = end.get_offset() - start.get_offset();
 
         const sound = SoundServer.getDefault();
+        const options = this._calculateKeystrokePitch();
```
ptomato: I would move this inside the `if (deletedLength === 1)` condition, so that we are only affecting the normal use of the backspace or delete keys, with no selection.

src/codeview.js
```diff
@@ -71,24 +72,42 @@ const Buffer = GObject.registerClass(class Buffer extends GtkSource.Buffer {
...
+
         if (deletedLength === 1)
-            sound.play('codeview/keypress/delete');
+            sound.playFull('codeview/keypress/delete', options);
         else
-            sound.play('codeview/keypress/delete-selection');
+            sound.playFull('codeview/keypress/delete-selection', options);
```
ptomato: ...and change this line back to what it was.

ptomato (changes requested):
One small change requested. I think it's closer to what David was describing, if we only vary the backspace/delete key sound that is played during a normal backspace/delete key press, not when there is a selection. (Because when you stop typing and select a region with the mouse as I expect most players will, then you exceed the pitch reset delay anyway.)

Otherwise looking and working well.

cfoch:
I agree with the comments. I was thinking the same but not sure if you would be thinking the same, too.

ptomato (approved):
:+1:

PR #128, opened 2019-02-01
window: Play toolbox background music

ptomato:
This plays the toolbox background music while the toolbox exists. Since
the sound server will now only play one background music at a time, we
can simply start it and stop it when a toolbox is created or shut down
respectively, and gains or loses toplevel focus.

https://phabricator.endlessm.com/T25022

### DIFF
--- a/src/window.js
+++ b/src/window.js
@@ -2,6 +2,7 @@
 
 const {Gio, GLib, GObject, Gtk, HackToolbox} = imports.gi;
 const {Lockscreen} = imports.lockscreen;
+const SoundServer = imports.soundServer;
 
 var ToolboxWindow = GObject.registerClass({
     Properties: {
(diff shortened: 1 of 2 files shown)


src/window.js
```diff
@@ -68,6 +79,33 @@ var ToolboxWindow = GObject.registerClass({
         });
     }
 
+    _startMusic() {
```
cfoch: Instead of implementing _startMusic and _stopMusic, could you use the SoundItem class from https://github.com/endlessm/gnome-shell/blob/master/js/misc/soundServer.js#L34-L83 ?

cfoch:
Besides my above comment. It does not seems like a good user experience. I am trying this with Fizzics. If the application is flipped and you unfocus it, the front background sound from Fizzics is resumed back. That is not good in my opinion.

ptomato:
I agree with your comment. So we should add something to stop the front-side music after all, like we have in the System app. I'll open a separate pull request for hack-toy-apps. This one is ready for review again.

cfoch:
There is a problem that I don't know if should be acceptable. If you have flipped the game, when you drag the window from the window border it seems that the current window loses the focus and so the Fizzics frontside sound is resumed. Then you drop the window and the backside sound is resumed.

It seems there is a bug in the hack-sound-server also... try flipping multiple times and close the window when you are on Fizzics. The Fizzics frontside sound will be kept playing besides even when this was unrefd. This is actually released when on EOS it seems instead after a fade out of 1s.
It seems that this is not fading out properly.

cfoch:
I think that adding
```
        if self._stop_loop:
            return
```
just after the _play method in the hack-sound-server will solve the problem

PR #130, opened 2019-02-14
toolbox: Add new action to init the toolbox

danigm:
The toolbox is being lazy loaded when the user click the FtH button. If
the toolbox is a bit complex, this loading can take some time and the
feedback for the user is odd because the interface freezes after the
click.

To try to *improve* the flipping time, we can preload the hack toolbox
for the app before the user clicks the FtH.

This patch adds a new action to create the toolbox widget so we can call
this from the shell and we can do the process transparent for the user.

https://phabricator.endlessm.com/T25093

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -58,6 +58,45 @@ var HackToolboxApplication = GObject.registerClass(class extends Gtk.Application
         });
         flip.connect('activate', this._onFlip.bind(this));
         this.add_action(flip);
+
+        const init = new Gio.SimpleAction({
+            name: 'init',
+            parameter_type: new GLib.VariantType('(ss)'),
+        });
+        init.connect('activate', this._initToolbox.bind(this));
+        this.add_action(init);
+    }
+
+    _initToolbox(action, parameterVariant) {
+        this.hold();
+        try {
+            const unpacked = parameterVariant.deep_unpack();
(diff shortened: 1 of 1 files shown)


src/app.js
```diff
@@ -58,6 +58,42 @@ var HackToolboxApplication = GObject.registerClass(class extends Gtk.Application
...
+        const init = new Gio.SimpleAction({
+            name: 'init',
+            parameter_type: new GLib.VariantType('(ss)'),
+        });
+        init.connect('activate', this._initToolbox.bind(this));
+        this.add_action(init);
+    }
+
+    _initToolbox(action, parameterVariant) {
+        this.hold();
```
ptomato, line 71: It's good practice when pairing hold/release calls to do something like this:
```js
this.hold();
try {
    doStuff();
} finally {
    this.release();
}
```
This avoids accidentally unpairing the calls in case there is an exception.

(I realize the existing code isn't doing that, but it would be good to do it in new code.)

ptomato (changes requested):
This makes a lot of sense to me. I have one comment about improving the robustness, otherwise this is good to merge.

src/app.js
danigm: I've added the try-finally to the new code.

ptomato (approved):
:+1:

PR #135, opened 2019-02-25
System app: Add trap sequence in wobbly lockscreen

manuq:
When the wobbly windows lockscreen is clicked, a trap sequence is
presented to the player, if running a specific quest. This sequence is
driven by the quest using the game state server as communication
channel.

If the player has the key needed to unlock the panel, the 'has-key'
background will be displayed as usual, but the player won't be able to
open it when clicked. Instead, game state 'tried' is set to the lock
for a few seconds. A quest can monitor this change to act
accordingly. There is a quest (Investigation) that presents a message,
telling the player to go away. And there is a quest (Breaking In) that
runs the trap sequence.

https://phabricator.endlessm.com/T25505


joaquimrocha (approved):
I am probably not the best candidate to review this PR. But @manuq asked me too and it LGTM, so approving it from my side.

manuq:
I made eslint pass :P

manuq:
Thanks @joaquimrocha for reviewing . @ptomato I'm merging this in the current state for today's playthrough, and I can continue working on this in follow up commits to improve the code add the final assets.

ptomato:
OK, I'll give this an after-the-fact review, thanks for making the extra effort to get it into the Monday build!

src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -0,0 +1,51 @@
...
+                GLib.Source.remove(this._timeout);
+            }
+        });
+    }
+
+    _updateBackground() {
+        super._updateBackground();
+        const assetsPath = this.getAssetsPath();
+        if (this._manager.getTrapSequence(this.lock) === 0)
+            this._playbin.background = `file://${assetsPath}/trap0`;
```
ptomato, line 26: This is a pre-existing situation, but do you know why we didn't include the lockscreen assets in the GResource and instead are installing them to `pkgdatadir`?

src/lockscreen.js
```diff
@@ -108,6 +108,25 @@ var Lockscreen = GObject.registerClass({
...
+                // All the required assets are here, let's use this path for
+                // the background
+                return path;
+            }
+        }
+
+        return defaultPath;
+    }
+
+    // FIXME refactor to use _getAssetsPath
```
ptomato, line 129: If you have more time to follow up, I would suggest doing this refactor, it seems like it should be quite straightforward.

src/locksManager.js
```diff
@@ -54,4 +54,20 @@ var LocksManager = GObject.registerClass({
...
     }
+
+    getTrapSequence(lock) {
+        return this._proxy.getDictValueSync(lock, 'trap_sequence');
+    }
+
+    setLockTried(lock, tried) {
+        try {
+            const [variant] = this._proxy.GetSync(lock);
+            const dict = new GLib.VariantDict(variant);
```
ptomato, line 65: I am pretty sure this can be done with `new GLib.Variant('a{sv}', {tried: new GLib.Variant('b', tried)});`

src/locksManager.js
```diff
@@ -54,4 +54,20 @@ var LocksManager = GObject.registerClass({
...
+            dict.insert_value('tried', new GLib.Variant('b', tried));
+            this._proxy.SetSync(lock, dict.end());
+        } catch (error) {
+            const variant2 = new GLib.Variant('a{sb}', {tried: true});
```
ptomato, line 69: This needs to be `{tried}` as otherwise the `tried` parameter given to the function is ignored, I think?

Also, a pre-existing problem, but I would really like to understand and fix where this discrepancy in variant types with the game state server is coming from, it caused problems with the hackdex corruption. I'll open an issue for that.

src/locksManager.js
```diff
@@ -54,4 +54,20 @@ var LocksManager = GObject.registerClass({
...
+    getTrapSequence(lock) {
+        return this._proxy.getDictValueSync(lock, 'trap_sequence');
+    }
+
+    setLockTried(lock, tried) {
+        try {
+            const [variant] = this._proxy.GetSync(lock);
+            const dict = new GLib.VariantDict(variant);
+            dict.insert_value('tried', new GLib.Variant('b', tried));
+            this._proxy.SetSync(lock, dict.end());
```
ptomato, line 67: Do we need to have this be sync? Can we just fire-and-forget an async operation instead?

data/OperatingSystemApp/controlpanel.ui
```diff
@@ -136,7 +136,7 @@
       </packing>
     </child>
     <child>
-      <object class="Lockscreen" id="wobblyLock">
+      <object class="WobblyLockscreen" id="wobblyLock">
```
ptomato, line 139: This will make this file impossible to load in Glade unless you add it to the Glade catalog too.

src/locksManager.js
```diff
@@ -54,4 +54,20 @@ var LocksManager = GObject.registerClass({
         const variant = new GLib.Variant('a{sb}', {locked: false});
         this._proxy.SetSync(lock, variant);
     }
+
+    getTrapSequence(lock) {
+        return this._proxy.getDictValueSync(lock, 'trap_sequence');
```
ptomato, line 59: I would try to make this async if it's feasible (it may not be.)

Additionally, we call it three times in a row, so at least perform only one sync call and save the value.

ptomato (commented):
A few comments that can be cleared up with some subsequent fixes, but overall it was the right decision to merge this as-is.

data/OperatingSystemApp/controlpanel.ui
manuq: Ouch! Good point.

src/OperatingSystemApp/wobblyLockscreen.js
manuq: No idea. In the clubhouse we also don't use the GResource except for the CSS, the UI files, and some background assets. But not for characters or items assets.

src/locksManager.js
manuq: I will check why this is not async. As I'm new to this codebase I repeated what the silbling methods do. Good point on saving the value.

src/locksManager.js
manuq: I did the same as similar methods in this module. So I will check if this can be async.

src/locksManager.js
manuq: Ouch! good catch.

src/lockscreen.js
manuq: Yes, I plan to do it. I wanted to get videos in first, as video playback may change how I'm getting the paths to these assets.

src/locksManager.js
manuq: I tried your suggestion using a new variant and it didn't work. What this does is update an existing lock. I did it in the same way as setUsed above in the file. So for example:
```
  "lock.OperatingSystemApp.2" : {
    "locked" : false
  }
```
Becomes:
```
  "lock.OperatingSystemApp.2" : {
    "locked" : false,
    "tried" : true
  }
```

manuq (commented):
Sorry I thought all my comment were sent but they were "pending".

src/locksManager.js
ptomato: Oh, I see, it's the equivalent of `dict.update()` in Python. Never mind my comment then

PR #137, opened 2019-02-28
T25505 videos

manuq:
https://phabricator.endlessm.com/T25505


src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -21,7 +21,8 @@ var WobblyLockscreen = GObject.registerClass({
 
     _updateBackground() {
         super._updateBackground();
-        const assetsPath = this.getAssetsPath();
+        // eslint-disable-next-line comma-spacing
+        const [assetsPath,,] = this.getAssetsPath();
```
ptomato: Here you can just do `const [assetsPath] = ...` to get only the first returned element

src/lockscreen.js
```diff
@@ -156,13 +137,15 @@ var Lockscreen = GObject.registerClass({
...
+        return [assetsPath, videoPath, assetsHasKey];
+    }
 
-        this._playbin.hasKey = this._key && this._manager.hasKey(this._key);
+    _updateBackground() {
+        const [assetsPath, videoPath, assetsHasKey] = this.getAssetsPath();
 
+        this._openURI = `file://${videoPath}/open.webm`;
+
+        this._playbin.hasKey = this._key && this._manager.hasKey(this._key);
```
ptomato, line 145: This looks a bit unclear, unless the `_manager.hasKey()` method has a side effect, it looks like it's not doing anything?

ptomato (changes requested):

src/lockscreen.js
manuq: Yes this was already there, I refactored without changing it. I think it's wrong. I will change it.

ptomato (approved):

PR #138, opened 2019-02-28
LightSpeed: Add lock.lightspeed.1 assets

xjuan:
https://phabricator.endlessm.com/T25596


xjuan:
Did not tested it, but it should work

PR #139, opened 2019-02-28
data: Replace placeholder assets for the trapped sequence

manuq:
https://phabricator.endlessm.com/T25630#683909


ptomato:
Looks like some of the new assets gained a `+x` bit in their mode, otherwise go ahead and merge this.

manuq:
@ptomato yeah I noticed that too and fixed it. Also I got final assets for the lock states "no key" and "has key", in which the lock glows.

PR #141, opened 2019-02-28
TEMPORARY - LightSpeed: Hack nextLevel instead of currentLevel

ptomato:
https://phabricator.endlessm.com/T25618


ptomato:
I will merge this workaround for Monday. I've been using it locally and it unblocks the quests, although it breaks the experience when you open the app outside of a quest.

PR #148, opened 2019-03-08
Ability to progressively reveal topics from game state

ptomato:
https://phabricator.endlessm.com/T25677

### DIFF
--- a/src/LightSpeed/toolbox.js
+++ b/src/LightSpeed/toolbox.js
@@ -1,11 +1,13 @@
 /* exported LSToolbox */
 
 const {Gio, GLib, GObject, Gtk} = imports.gi;
+const Signals = imports.signals;
 
-const {Toolbox} = imports.toolbox;
+const GameState = imports.gameState;
 const {LSCombinedTopic} = imports.LightSpeed.controlpanel;
 const {LSGlobalModel} = imports.LightSpeed.globalParams;
 const {LSUserFunction} = imports.LightSpeed.userFunction;
+const {Toolbox} = imports.toolbox;
 
 var LSToolbox = GObject.registerClass(class LSToolbox extends Toolbox {
     _init(appId, props = {}) {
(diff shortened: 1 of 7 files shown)


src/LightSpeed/toolbox.js
```diff
@@ -18,33 +20,46 @@ var LSToolbox = GObject.registerClass(class LSToolbox extends Toolbox {
 
         this._combinedTopic = new LSCombinedTopic();
         this._combinedTopic.bindGlobal(this._global);
-        this.addTopic('Game', 'rocket-symbolic', this._combinedTopic);
+        this.addTopic('game', 'Game', 'rocket-symbolic', this._combinedTopic);
 
         this._spawnEnemyTopic = new LSUserFunction('spawnEnemy');
         this._spawnEnemyTopic.bindGlobal(this._global);
-        this.addTopic('Enemies', 'astronaut-symbolic', this._spawnEnemyTopic);
+        this.addTopic('spawnEnemy', 'Enemies', 'astronaut-symbolic',
```
MegFord, line 27: Do you want me to wait to merge this until you have the correct icon for Enemies or do you want to put that in a different PR?

src/LightSpeed/toolbox.js
```diff
@@ -67,4 +82,32 @@ var LSToolbox = GObject.registerClass(class LSToolbox extends Toolbox {
         this._combinedTopic.reset();
         // The user functions are reset by the LightSpeed app already
     }
+
+    async _showTopicsInitially() {
+        const topics = ['spawnEnemy', 'updateAsteroid', 'updateSpinner',
```
MegFord, line 87: Should 'game' be included here?

src/LightSpeed/toolbox.js
ptomato: No, I think we can replace the icons separately when the assets are delivered.

src/LightSpeed/toolbox.js
ptomato: Yeah, good question. I originally didn't include it because that topic is always shown unconditionally, and adding it here would require us to set its corresponding game state key somewhere, otherwise it wouldn't show up. However, that does make it so that we couldn't hide that topic in the future unless we made that change here.

Your call as the reviewer, I guess :smile: I can implement it either way.

MegFord: I see! Thanks for the explanation. In that case I'd leave the refactor until it's necessary.

PR #151, opened 2019-03-09
toolbox: Replace background asset

ptomato:
This asset reaches all the way to the bottom of the screen on the
officially supported Hack resolution.

https://phabricator.endlessm.com/T25755


ptomato:
Merging this unreviewed, to improve things for the Monday playthrough, as it's a simple asset replacement. I tested it at 1080px resolution, as well as a smaller resolution with 2x scale.

PR #152, opened 2019-03-09
Lightspeed: Replace icon assets for topics

ptomato:
Converted from the original assets linked in the Phabricator task, by
resizing them to 16x16; removing the background rectangles; changing the
fills to #bebebe, the chroma key color; and running object-to-path on
the <circle>s in spawn-symbolic.svg because they otherwise won't get
recolored due to a recently fixed GTK bug.

https://phabricator.endlessm.com/T25756


ptomato:
Likewise, merging this without review in order for expediency Monday. It switches a few assets, and renames one (the rename requires one line of code change.) I tested it before merging.

PR #154, opened 2019-03-12
Duplicate window bugfix

ptomato:
https://phabricator.endlessm.com/T25710


src/app.js
```diff
@@ -78,6 +78,9 @@ var HackToolboxApplication = GObject.registerClass(class extends Gtk.Application
             if (!this._windows[appId])
                 this._windows[appId] = {};
 
+            if (this._windows[appId][windowId])
+                return;
```
cfoch, line 82: I was doing the same :-)

PR #155, opened 2019-03-13
codeview: Work around mozjs52 Reflect.parse() bug

ptomato:
The location information for the function body is incorrect. This bug
has been fixed already in mozjs60, but we are stuck on this runtime for
now and cannot update to >= GNOME 3.30.

Instead of using the function body ending information, use the function
declaration ending information. In mozjs60, these end at the same place,
so I assume that the function declaration is not meant to include the
closing brace.

https://phabricator.endlessm.com/T25802


manuq (approved):
Good hack!

PR #158, opened 2019-03-15
T25679

MegFord:


data/OperatingSystemApp/icons/16x16/actions/cursor-symbolic.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="70px" height="70px" viewBox="0 0 70 70" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
ptomato: These are going to need a bit of editing to get them into the format that GTK expects for symbolic SVG icons ... they have to be on a 16x16 pixel canvas, and all the fills should be `#bebebe` - this is a chroma key color that allows recoloring the icon in hover states, for example. (Right now it's not noticeable because we recolor the icons to `#ffffff` anyway in the CSS)

I had the same problem with the Lightspeed topic icons, for expediency I've just been doing the edits myself in Inkscape instead of sending the icons back for a round of revision. But I'm pretty sure I wrote these requirements down somewhere and shared it with the designers. After the release I should dig them up and ask them to keep the requirements in mind.

data/glade/toolbox.glade.xml
```diff
@@ -65,8 +65,10 @@
       </properties>
     </glade-widget-class>
 
-    <glade-widget-class name="Section" generic-name="section"
-                        template="../section.ui" title="Section">
+    <glade-widget-class name="Section"
```
ptomato: Unrelated change?

src/OperatingSystemApp/toolbox.js
```diff
@@ -1,28 +1,42 @@
...
     _init(appId, props = {}) {
         super._init(appId, props);
+
+        /* Setup icons path */
+        var theme = Gtk.IconTheme.get_default();
+        theme.add_resource_path('/com/endlessm/HackToolbox/OperatingSystemApp/icons');
+
         this._controlPanel = new OSControlPanel({visible: true});
-        this.addTopic('main', 'Tools', 'preferences-other-symbolic', this._controlPanel);
+        this.addTopic('main', 'Tools', 'cursor-symbolic', this._controlPanel);
```
ptomato: Should we rename this topic to "Cursor" now instead of "Tools"? In the storyboard it's labeled "Background" but I don't think that's correct either.

src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -2,12 +2,11 @@
 
 const {GLib, GObject} = imports.gi;
 
-const {Lockscreen} = imports.lockscreen;
 const SoundServer = imports.soundServer;
+const {Lockscreen} = imports.lockscreen;
```
ptomato: I'd suggest moving this line back to where it was, we try (although don't always succeed) to keep diffs small and imports alphabetized

src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -2,12 +2,11 @@
...
 
 var WobblyLockscreen = GObject.registerClass({
     GTypeName: 'WobblyLockscreen',
-    CssName: 'wobblylockscreen',
```
ptomato: What's the reason to remove this? (I'm sure there's a good one, but I don't understand it :smile:)

src/OperatingSystemApp/controlpanel.js
```diff
@@ -47,9 +38,8 @@ var OSControlPanel = GObject.registerClass({
         // the old design
         this._codeview.minContentHeight = 424;
 
-        /* Create data models for editable properties */
+        /* Create data model for editable properties */
         this._cursor = new OSCursorModel();
-        this._wobbly = new OSWobblyModel();
```
ptomato, line 52: Fixing this is not required for merging the PR, but one improvement to make here in the future might be to move the models to `toolbox.js` and add a `bindModel()` method on each of the topic control panels, so the toplevel toolbox `_init()` does something approximately like
```js
this._cursorModel = new OSCursorModel();
this._cursorTopic.bindModel(this._cursorModel);
```
I believe that's how it's done in the other toolboxes.

Since we're short on time, it's not necessary now, but just mentioning this to give you context on the existing patterns.

ptomato (changes requested):
Thanks! So far this is working great, I have yet to test it with the trapped sequence. I will do that first thing after lunch, and try to pay attention particularly to the issues you mentioned in Slack.

Couple of miscellaneous comments:
- I would set the `valign` of the cursor Section to START instead of FILL, since there's no longer another section below it.
- If you put a link to the Phabricator ticket in your commit message, then Phabricator will auto-comment on the ticket when the PR is opened and closed. We also generally try to put a 1 or 2 sentence summary of how the commit fixes the ticket, in this case it could be something like "This moves the wobbly windows controls into their own topic, and adds a second code view to that topic, similar to the code view in the cursor topic."
- These three commits should preferably be squashed into one.

data/glade/toolbox.glade.xml
MegFord: Sorry I should have reverted that. I was messing around with that because I couldn't get Glade to load the section class.

src/OperatingSystemApp/wobblyLockscreen.js
MegFord: Nah, in general I don't have good reasons for things. Like the mistake above, this is something I did when I was trying to figure out why the styling was incorrect on the Section asset. I'll revert :)

ptomato (approved):
All working well, thanks! I'll open the follow up tickets that we discussed on Slack.

PR #162, opened 2019-03-19
Fake Saniel dialog box

ptomato:
https://phabricator.endlessm.com/T25690


src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -116,4 +122,48 @@ var WobblyLockscreen = GObject.registerClass({
...
+                halign: Gtk.Align.START,
+                label: 'Okay',
+                valign: Gtk.Align.END,
+            });
+            const saniel = new Gtk.Frame({
+                halign: Gtk.Align.END,
+                name: 'saniel-head',
+                shadowType: Gtk.ShadowType.NONE,
+                valign: Gtk.Align.FILL,
+            });
```
joaquimrocha, line 156: I thought you would copy the UI file + CSS rules we already have in the Clubhouse for the in-window dialog. But we already have this so let's keep it.

src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -123,7 +125,47 @@ var WobblyLockscreen = GObject.registerClass({
...
+        }
+
+        let breakingInActive = false;
+        try {
+            const variant = await gss.Get('quest.BreakingIn');
+            breakingInActive = getActive(variant);
+        } catch (e) {
+            if (!isGameStateKeyError(e))
+                throw e;
+            breakingInActive = false;
```
joaquimrocha: I guess either we delete this entry here or we set the `investigationActive` variable above too when there's an error.
The former makes more sense to me because the variable shouldn't be set if there's an exception in this logic.

src/OperatingSystemApp/wobblyLockscreen.js
```diff
@@ -123,7 +125,47 @@ var WobblyLockscreen = GObject.registerClass({
...
+        const gss = GameState.getDefault();
+
+        let investigationActive = false;
+        try {
+            const variant = await gss.Get('quest.Investigation');
+            investigationActive = getActive(variant);
+        } catch (e) {
+            if (!isGameStateKeyError(e))
+                throw e;
+        }
```
joaquimrocha: I think that the GSS is not the best place for dealing with this. The goal of the GSS is configuration and this is more like a temporary state. More importantly, it has the issues that if there's a bug/crash, the `active` key may be left in the GSS, and requires managing this key's life in the Clubhouse otherwise.

I think it's also interesting for the Clubhouse to share what is the currently running quest with whoever is interested. So I will add a D-Bus property for that and try to add a commit here with that change too.

joaquimrocha (dismissed):

joaquimrocha:
@ptomato I have pushed a couple of branches to this PR. One of them reverts the GSS related commit, and it's a fixup one, so upon squashing the reverted commit should rather be removed.

The other commit is for checking a new DBus property of the Clubhouse that I have included in https://github.com/endlessm/clubhouse/pull/422 .

I see the CI is failing but I don't think it's in the changes I added (I will let you decide what to do with those since you should review my commit too).

src/OperatingSystemApp/wobblyLockscreen.js
ptomato: You're right, I had forgotten about the in-window dialog, I just looked at the one in gnome-shell and saw some incompatibilities between ST and GTK CSS.

src/OperatingSystemApp/wobblyLockscreen.js
ptomato: The exception here is that the key doesn't exist in the game state, which happens if you haven't reached that part of the episode yet. Any other exceptions (`!isGameStateKeyError`) are re-thrown in the line above, so the variable isn't set.

joaquimrocha: But you do declare the variable before the `try`, so this shouldn't be needed AFAIK. Anyway, in the solution I am proposing this code is no longer needed in any case :slightly_smiling_face:

ptomato: True, I don't need to set the variable a second time here. I misunderstood "delete this entry here", thought you were saying I needed to unset the variable.

joaquimrocha:
I've just found out I had forgotten to include the `clubhouse.js` file... :roll_eyes: 
Commit updated...

src/clubhouse.js
```diff
@@ -0,0 +1,33 @@
+const {Gio} = imports.gi;
+
+const ClubhouseIface =
+'<node> \
```
ptomato: For future reference, use a backtick string for this, then you don't have to escape the newlines.

ptomato (commented):
Yes, works for me. I will test it this afternoon and then merge it.

The linter errors are because of mixed tabs and spaces... It's no problem for me to change it, but you may want to double-check your editor configuration.

ptomato:
I spent some time testing this. I'll merge it now.

PR #163, opened 2019-03-19
Fix double sounds

danigm:
For fizzics and lightspeed, the buttons are linked to modify the code on change, this is done connecting the model property notify signal and this model is not unref when the widget is destroyed so if you close and open again fizzics, every click will play two sounds and we'll have a traceback in the log because the callback is trying to update a destroyed widget.

This PR fixes this problem disconnecting all handlers from the model on widget destroy.

### DIFF
--- a/src/LightSpeed/userFunction.js
+++ b/src/LightSpeed/userFunction.js
@@ -126,12 +126,20 @@ var LSUserFunction = GObject.registerClass({
         this.get_style_context().add_class('user-function');
     }
 
-    bindGlobal(model) {
-        this._global = model;
-        this._globalNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
+    bindModel(model) {
+        this._model = model;
+        this._modelNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
 
         this._currentLevel = model.currentLevel;
-        this._bindModel(this._global.getModel(this._currentLevel));
+        this._bindModel(this._model.getModel(this._currentLevel));
+    }
+
+    unbindModel() {
+        this._unbindLocalModel();
+        if (this._model && this._modelNotifyHandler) {
+            this._model.disconnect(this._modelNotifyHandler);
+            this._modelNotifyHandler = 0;
+        }
     }
 
     _onGlobalNotify() {
@@ -141,11 +149,15 @@ var LSUserFunction = GObject.registerClass({
         this._bindModel(this._global.getModel(this._currentLevel));
     }
 
-    _bindModel(model) {
-        if (this._model)
-            this._model = null;
+    _unbindLocalModel() {
+        this._localModel = null;
+    }
 
-        this._model = model;
+    _bindLocalModel(model) {
+        if (this._localModel)
+            this._localModel = null;
+
+        this._localModel = model;
 
         const {name, args, modelProp} = USER_FUNCTIONS[this._codeview.userFunction];
         this._codeview.text = `function ${name}(${args.join(', ')}) {
@@ -201,6 +213,6 @@ ${code}
         this._codeview.setCompileResults([]);
 
         const funcBody = this._codeview.getFunctionBody(name);
-        this._model[modelProp] = funcBody;
+        this._localModel[modelProp] = funcBody;
     }
 });
--- a/src/LightSpeed/controlpanel.js
+++ b/src/LightSpeed/controlpanel.js
@@ -36,35 +36,47 @@ var LSCombinedTopic = GObject.registerClass({
         this._variablesCodeview.connect('should-compile', this._compile.bind(this));
     }
 
-    bindGlobal(model) {
-        this._global = model;
-        this._globalNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
+    bindModel(model) {
+        this._model = model;
+        this._modelNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
 
         this._currentLevel = model.currentLevel;
-        this._bindModel(this._global.getModel(this._currentLevel));
+        this._bindLocalModel(this._model.getModel(this._currentLevel));
+    }
+
+    unbindModel() {
+        this._unbindLocalModel();
+        if (this._model && this._modelNotifyHandler) {
+            this._model.disconnect(this._modelNotifyHandler);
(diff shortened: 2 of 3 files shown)


src/Fizzics/level2.js
```diff
@@ -30,6 +30,13 @@ var FizzicsLevel2 = GObject.registerClass({
             this._compile();
         });
         this._content.add(this._codeview);
+
+        this.connect('destroy', () => {
+            if (this._model && this._notifyHandler) {
+                this._model.disconnect(this._notifyHandler);
+                this._notifyHandler = null;
```
joaquimrocha: Not that it matters a lot here, but the signal handlers are integers, so this should check if they're `0` and reset it as such IMO. Same comment for all the rest of the uses.

joaquimrocha: We should also check if adding an `unbindModel` (empty) method to the base class, and calling it in there in its `shutdown` method. So instead of each widget having to remember to connect to the destroy signal, they'd just have to disconnect whatever they connected during the `bindModel`, if that makes sense (and works).

joaquimrocha (changes requested):
Left a couple of comments.

src/Fizzics/level2.js
danigm: > Not that it matters a lot here, but the signal handlers are integers, so this should check if they're `0` and reset it as such IMO. Same comment for all the rest of the uses.

As far as I know, the handler can't be 0, this is the error that I got when I use an invalid handler_id:
```
g_signal_handler_disconnect: assertion 'handler_id > 0' failed
```

joaquimrocha: Of course. The handler is an unsigned long ( https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-handler-disconnect ).

My point is that you're treating the handler here like an object (checking/reassigning to `null`) but it's a ulong, so it should be checked and reset as `0`.

danigm: Okay, now I undertstand this and I've changed it.

src/Fizzics/level2.js
```diff
@@ -328,4 +328,11 @@ var FizzicsLevel2 = GObject.registerClass({
             this._onNotify.bind(this));
         this._regenerateCode();
     }
+
+    unbindModel() {
+        if (this._model && this._notifyHandler) {
```
joaquimrocha: Nitpick: I'd prefer an explicit `this._notifyHandler != 0`; same with the other calls, but reminder: I am not an expert on js idioms.

src/Fizzics/toolbox.js
```diff
@@ -31,4 +31,9 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
         win.get_style_context().add_class('Fizzics');
         this._controlPanel.bindWindow(win);
     }
+
+    shutdown() {
+        super.shutdown();
```
joaquimrocha: I thought that calling `super()` would do this, but I am not a Js expert... @ptomato Maybe knows more about the correct way here.

src/Fizzics/level2.js
ptomato: JS isn't typed, so it doesn't matter that the handler ID is otherwise an integer. Using `null` for a not-yet-assigned class member is a common idiom and the reason why we do it is so that the programmer only has to remember to check `!== null` or `=== null` instead of worrying about any of JS's weird type coercions that may apply.

src/LightSpeed/userFunction.js
```diff
@@ -139,11 +147,15 @@ var LSUserFunction = GObject.registerClass({
...
-    _bindModel(model) {
-        if (this._model)
-            this._model = null;
+    _unbindLocalModel() {
+        this._localModel = null;
+    }
+
+    _bindLocalModel(model) {
+        if (this._localModel)
+            this._localModel = null;
```
joaquimrocha, line 158: Why do you set this do `null` if you're already assigning it below? I see the code was already there, but I don't understand why that was needed eitheer.

joaquimrocha (approved):
I don't think there's anything critical but I left a couple of clean-up comments, at this point I'd prefer if @ptomato verified the PR too :)

src/Fizzics/toolbox.js
ptomato: Unlike Python, `super(whatever)` syntax is only in the constructor. In other methods you call `super.methodName(whatever)`. So this code is correct. Not sure why the JS committee did it that way.

ptomato (commented):
I find it difficult to estimate the risk profile of this PR due to the big rename in the Lightspeed commit (`global` -> `model`, `model` -> `localModel`.) On the one hand I'd like to avoid the rename if possible, so close to the release, because if one instance was forgotten then it could easily be overlooked, and it would be ambiguous. (And from a "naming things is hard" perspective, it shouldn't be called "local" - it's a per-level model, it's just as remote and non-local as the global model :smile:)

On the other hand, this commit is what you've tested with, so if we change that then we have less testing in total.

I think what I'll do is merge the Fizzics and Framework commits from this branch, and rebase the Lightspeed commit and make an alternative PR without the rename so we can see if the diff gets substantially smaller and/or easier to review in a way that seems to reduce the risk. If that's the case, then I think that should outweigh the testing risk.

----

I think we need the same thing implemented for the System app toolbox as well. It's not strictly necessary for the HackUnlock toolbox, since players never start another instance of HackUnlock, but we should probably do the same thing there for completeness. (If we do it for HackUnlock, then we should merge that after the freeze, though.)

src/LightSpeed/userFunction.js
```diff
@@ -126,12 +126,20 @@ var LSUserFunction = GObject.registerClass({
...
-    bindGlobal(model) {
-        this._global = model;
-        this._globalNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
+    bindModel(model) {
+        this._model = model;
+        this._modelNotifyHandler = model.connect('notify', this._onGlobalNotify.bind(this));
 
         this._currentLevel = model.currentLevel;
-        this._bindModel(this._global.getModel(this._currentLevel));
+        this._bindModel(this._model.getModel(this._currentLevel));
```
ptomato, line 134: This would need to be `_bindLocalModel` if we went with the rename

PR #165, opened 2019-03-19
tools: Add simple way to build a local flatpak

manuq:
This is a bare copy of the clubhouse tool, with modified filenames for
the JSON manifests.

### DIFF
--- a/tools/build-local-flatpak.sh
+++ b/tools/build-local-flatpak.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# Convenience script to build a Flatpak bundle with the local contents (the original manifest)
+# builds from git so changes need to be committed.
+#
+# Copyright (c) 2018 Endless Mobile Inc.
+#
+# Author: Joaquim Rocha <jrocha@endlessm.com>
+#
+
+set -e
+source_dir="$(git rev-parse --show-toplevel)"
+
+pushd "$source_dir"
+
+# If the GIT_CLONE_BRANCH var is not defined, we replace it so that it'll in fact set the
+# git branch to "" but add also "type": "dir" so it builds from the current directory.
+# This is hacky but until we need to keep git as the source type, then it's the less intrusive.
+GIT_CLONE_BRANCH=${GIT_CLONE_BRANCH:-'", "type": "dir'}
+REPO=${REPO:-repo}
+BRANCH=${BRANCH:-master}
+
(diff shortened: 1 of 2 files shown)


tools/build-local-flatpak.sh
```diff
@@ -0,0 +1,32 @@
...
+GIT_CLONE_BRANCH=${GIT_CLONE_BRANCH:-'", "type": "dir'}
+REPO=${REPO:-repo}
+BRANCH=${BRANCH:-master}
+
+sed -e "s|@GIT_CLONE_BRANCH@|${GIT_CLONE_BRANCH}|g" \
+    -e "s|@BRANCH@|${BRANCH}|g" \
+  com.endlessm.HackToolbox.json.in > com.endlessm.HackToolbox.json
+
+# Add any extra options from the user to the flatpak-builder command (e.g. --install)
+flatpak-builder build --user --force-clean com.endlessm.HackToolbox.json --repo=${REPO} $@ || ret=$?
```
ptomato, line 28: I would remove `--user` here and leave it up to the caller if they want to include, e.g. `--user --install` on the command line.

tools/build-local-flatpak.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+source_dir="$(git rev-parse --show-toplevel)"
+
+pushd "$source_dir"
+
+# If the GIT_CLONE_BRANCH var is not defined, we replace it so that it'll in fact set the
+# git branch to "" but add also "type": "dir" so it builds from the current directory.
+# This is hacky but until we need to keep git as the source type, then it's the less intrusive.
+GIT_CLONE_BRANCH=${GIT_CLONE_BRANCH:-'", "type": "dir'}
```
ptomato, line 19: I was not really a fan of this in the clubhouse repo either - it prints out a big warning whenever you build it. But it is harmless, I guess we should do it for now and keep working on a better solution in the meantime.

ptomato (changes requested):
This is not relevant to the PR, but I've found it a lot slower to use `--install` during development, than to have a local flatpak repo and call `flatpak update --no-deps --assumeyes com.endlessm.HackToolbox` as part of my build script. IIRC `--install` pops up annoying prompts for you to enter your password.

Anyway, we still have the old script, so people can choose which way works for them. The only thing I'd change here is remove `--user` from the script.

manuq:
Yeah Philip I'm all for these suggestions! I only wanted to unblock
@megford for now.

ptomato:
OK, let's go ahead with it then.

PR #166, opened 2019-03-19
LightSpeed: use consistent variable structure for positions

cosimoc:
The enemy position is recorded as enemy.position.x/y; do the same
for the player ship.

https://phabricator.endlessm.com/T25821


ptomato (approved):
Great, thanks!

PR #167, opened 2019-03-19
update the lock state if trap sequence is null and lock 2 is unlocked

MegFord:
https://phabricator.endlessm.com/T25894


ptomato (approved):
This seems fine, I tested what happens after the end of the episode and also the interaction with the fake Saniel dialog box that I just merged before this one.

PR #168, opened 2019-03-20
lightspeed: Disconnect notify on destroy

ptomato:
https://phabricator.endlessm.com/T25826

### DIFF
--- a/src/LightSpeed/userFunction.js
+++ b/src/LightSpeed/userFunction.js
@@ -134,6 +134,14 @@ var LSUserFunction = GObject.registerClass({
         this._bindModel(this._global.getModel(this._currentLevel));
     }
 
+    unbindGlobalModel() {
+        this._unbindLevelModel();
+        if (this._global && this._globalNotifyHandler) {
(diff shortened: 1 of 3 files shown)


ptomato:
Same as https://github.com/endlessm/hack-toolbox-app/pull/163, but I've removed the rename to make a smaller pull request. I've named the new methods `unbindGlobalModel` and `unbindLevelModel` to remove any ambiguity, even though I didn't rename the existing methods.

src/LightSpeed/userFunction.js
```diff
@@ -134,16 +134,28 @@ var LSUserFunction = GObject.registerClass({
         this._bindModel(this._global.getModel(this._currentLevel));
     }
 
+    unbindGlobalModel() {
+        this._unbindLevelModel();
+        if (this._global && this._globalNotifyHandler) {
+            this._global.disconnect(this._globalNotifyHandler);
+            this._globalNotifyHandler = null;
```
cosimoc, line 141: While we're touching this code, it would be nice to be consistent about either setting `this._global = null;` or not when it's unbound.

PR #169, opened 2019-03-20
lightspeed: Update userFunction code on model change

danigm:
https://phabricator.endlessm.com/T25897


joaquimrocha (approved):
LGTM! I think the convention is to use `Lightspeed` (capitalized) as the commit prefix, so maybe we should do that.

joaquimrocha:
:+1:

PR #170, opened 2019-03-20
LightSpeed: Block update code signal hander on editing

danigm:
As we're listening to every model change now to update the code we need
to block this signal handler during code editing to avoid the loop.

https://phabricator.endlessm.com/T25897

### DIFF
--- a/src/LightSpeed/userFunction.js
+++ b/src/LightSpeed/userFunction.js
@@ -174,8 +174,17 @@ ${this._model[modelProp]}
 }`;
     }
 
+    _updateCode(funcBody) {
+        const {modelProp} = USER_FUNCTIONS[this._codeview.userFunction];
+        // Block the normal notify handler that updates the code view, since we
+        // are propagating updates from the codeview to the GUI.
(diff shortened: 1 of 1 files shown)


src/LightSpeed/userFunction.js
```diff
@@ -174,8 +174,17 @@ ${this._model[modelProp]}
 }`;
     }
 
+    _updateCode(funcBody) {
+        const {modelProp} = USER_FUNCTIONS[this._codeview.userFunction];
+        // Block the normal notify handler that updates the code view, since we
+        // are propagating updates from the codeview to the GUI.
+        GObject.signal_handler_block(this._model, this._notifyHandler);
```
ptomato, line 181: For future reference, every time we have a paired block/unblock call, it's best to use a try/finally construction:
```js
GObject.signal_handler_block(...);
try {
    this._model[modelProp] = funcBody;
} finally {
    GObject.signal_handler_unblock(...);
}
```
This is because setting `modelProp` can call a JS setter function, which may throw.

PR #171, opened 2019-03-20
Lightspeed: Make a distinction between global and per-level functions

ptomato:
User functions can now be per-level (spawnEnemy) or global (the four
update functions.) We move the model properties for the latter from the
per-level model to the global model, and introduce code in
userFunction.js to connect up the right notify signal handlers for each
case.

https://phabricator.endlessm.com/T25722

### DIFF
--- a/src/LightSpeed/userFunction.js
+++ b/src/LightSpeed/userFunction.js
@@ -71,6 +71,7 @@ const USER_FUNCTIONS = {
         name: 'spawnEnemy',
         args: [],
         modelProp: 'spawnEnemyCode',
+        perLevel: true,
         getScope() {
             return Object.assign({}, COMMON_SPAWN_SCOPE, COMMON_SCOPE);
         },
@@ -80,6 +81,7 @@ const USER_FUNCTIONS = {
         args: [],
         modelProp: 'updateAsteroidCode',
         childWidget: '_updateAsteroidCodeview',
+        perLevel: false,
         getScope() {
             return Object.assign({}, COMMON_UPDATE_SCOPE, COMMON_SCOPE);
         },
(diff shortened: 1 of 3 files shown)


src/LightSpeed/userFunction.js
```diff
@@ -128,10 +133,18 @@ var LSUserFunction = GObject.registerClass({
...
+        if (perLevel) {
+            this._globalNotifyHandler = model.connect('notify',
+                this._onGlobalNotify.bind(this));
+            this._currentLevel = model.currentLevel;
+            this._bindModel(this._global.getModel(this._currentLevel));
+        } else {
+            this._globalNotifyHandler = model.connect(`notify::${modelProp}`,
+                this._setCode.bind(this));
+            this._setCode();
+        }
```
cosimoc, line 147: I think that it would result in a cleaner patch to always connect to `notify::${modelProp}` (or even `notify` I guess, not sure why the pre-existing code did not do that) and add the distinguishing logic to `onGlobalNotify()` instead.

cosimoc (commented):
Looks good, just a comment that could possibly simplify the implementation a bit.

src/LightSpeed/userFunction.js
ptomato: `modelProp` is on the per-level model in the pre-existing `perLevel` case, not on the global model. In the`perLevel` case, `_onGlobalNotify()` could actually connect to `notify::currentLevel` I guess, but I didn't want to change that right now.

ptomato:
If I understood correctly, I don't think I can implement that comment right now without more refactoring, so I'll put this one up for review again.

src/LightSpeed/userFunction.js
cosimoc: I have not tested this, so we should not take my patch right now, but I pushed this commit to a WIP branch to better illustrate what I meant: https://github.com/endlessm/hack-toolbox-app/commit/1fbca363a15ddd66909e2fbbae49dce76a240258

PR #174, opened 2019-03-26
Add appdata file

ptomato:
We must remove the Icon: line from the desktop file, as that icon does
not exist, and appstream-compose will complain about that.

I believe that sending metrics which include code typed in by the user
requires the "intense" rating for social-info in the OARS metadata.

https://phabricator.endlessm.com/T26020

### DIFF
--- a/data/com.endlessm.HackToolbox.appdata.xml
+++ b/data/com.endlessm.HackToolbox.appdata.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.endlessm.HackToolbox.desktop</id>
+  <metadata_license>CC0</metadata_license>
+  <project_license>LGPL-2.1</project_license>
+  <categories>
+    <category>Game</category>
+  </categories>
+  <name>Toolbox</name>
+  <summary>The other side of flip-to-hack</summary>
+
+  <description><p></p></description>
+
+  <url type="homepage">https://hack-computer.com/</url>
+
+  <releases>
+    <release date="2019-03-25" version="0.0.0"/>
+  </releases>
+
+  <content_rating type="oars-1.1">
+    <content_attribute id="social-info">intense</content_attribute>
+  </content_rating>
+
+  <update_contact>hack@endlessm.com</update_contact>
+</component>
--- a/data/com.endlessm.HackToolbox.desktop
+++ b/data/com.endlessm.HackToolbox.desktop
@@ -4,7 +4,6 @@ Name=Hack Toolbox
 Comment=Hack Toolbox
 Type=Application
 Exec=com.endlessm.HackToolbox
-Icon=hack-toolbox
 Categories=Development;
 DBusActivatable=true
 StartupNotify=true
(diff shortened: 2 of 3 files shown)


data/com.endlessm.HackToolbox.appdata.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.endlessm.HackToolbox.desktop</id>
+  <metadata_license>CC0</metadata_license>
+  <project_license>LGPLv2</project_license>
+  <categories>
+    <category>Game</category>
+  </categories>
+  <name>Toolbox</name>
+  <summary>The other side of flip-to-hack</summary>
```
cosimoc, line 10: I thought there was a "no display" option for the appdata XML as well, isn't there?

ptomato: Scanning through https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html, I don't see one; do you remember what it was called?

cosimoc: I see things like this in the appstream published for eos-apps, but I am copying @joaquimrocha here to be sure. Perhaps those are generated by appstream-compose by looking at the desktop file and the appdata.xml at the same time? Who knows!

```
    <vetos>
      <veto>NoDisplay=true</veto>
    </vetos>
```

joaquimrocha: @cosimoc That's true. The `vetos` work, and do get added automagically too. However, this and other similar projects don't get shown directly in g-s in any case because they need an icon (besides the name and summary) in order to be displayed.

data/com.endlessm.HackToolbox.appdata.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>com.endlessm.HackToolbox.desktop</id>
+  <metadata_license>CC0</metadata_license>
+  <project_license>LGPL-2.1</project_license>
+  <categories>
+    <category>Game</category>
+  </categories>
```
joaquimrocha, line 8: I don't think you need to add categories. This is not supposed to be visible to the uuser anyway.

data/com.endlessm.HackToolbox.appdata.xml
```diff
@@ -0,0 +1,25 @@
...
+  <name>Toolbox</name>
+  <summary>The other side of flip-to-hack</summary>
+
+  <description><p></p></description>
```
joaquimrocha, line 12: This is also not needed.

data/com.endlessm.HackToolbox.appdata.xml
```diff
@@ -0,0 +1,25 @@
...
+
+  <url type="homepage">https://hack-computer.com/</url>
+
+  <releases>
+    <release date="2019-03-25" version="0.0.0"/>
+  </releases>
```
joaquimrocha, line 18: I'd also not add this unless you intend to keep it up to date. It's of course harmless, so it's up to you.

joaquimrocha (approved):
I think the appdata file could be simplified, but none of this really represents an issue. So I am approving the changes.

joaquimrocha:
After talking to Simon, we both agree to merge this as is so we save time and are able to test ASAP.

PR #175, opened 2019-03-29
Add error indicator on Topic icon

MegFord:
https://phabricator.endlessm.com/T25678

### DIFF
--- a/src/LightSpeed/controlpanel.js
+++ b/src/LightSpeed/controlpanel.js
@@ -16,6 +16,11 @@ const VALID_SHIPS = ['spaceship', 'daemon', 'unicorn'];
 
 var LSCombinedTopic = GObject.registerClass({
     GTypeName: 'LSCombinedTopic',
+    Properties: {
+        'needs-attention': GObject.ParamSpec.boolean('needs-attention', 'Needs attention',
+            'Display an indicator on the button that it needs attention',
+            GObject.ParamFlags.READWRITE, false),
+    },
     Template: 'resource:///com/endlessm/HackToolbox/LightSpeed/panel.ui',
     InternalChildren: ['astronautSizeAdjustment', 'scoreTargetAdjustment',
         'shipAssetButton', 'shipSizeAdjustment', 'shipSpeedAdjustment',
@@ -120,6 +125,7 @@ var LSCombinedTopic = GObject.registerClass({
             if (!(e instanceof SyntaxError || e instanceof ReferenceError))
                 throw e;
             this._variablesCodeview.setCompileResultsFromException(e);
+            this.set_property('needs-attention', true);
             return;
         }
 
@@ -129,10 +135,12 @@ var LSCombinedTopic = GObject.registerClass({
         const errors = this._searchForErrors(scope);
         if (errors.length > 0) {
             this._variablesCodeview.setCompileResults(errors);
+            this.set_property('needs-attention', true);
             return;
         }
 
         this._variablesCodeview.setCompileResults([]);
+        this.set_property('needs-attention', false);
 
         // Block the normal notify handler that updates the code view, since we
         // are propagating updates from the codeview to the GUI. Instead,
@@ -205,6 +213,7 @@ var LSCombinedTopic = GObject.registerClass({
             SoundServer.getDefault().play('hack-toolbox/update-codeview');
             this._lastCodeviewSoundMicrosec = timeMicrosec;
         }
+        this.set_property('needs-attention', false);
     }
 
     _regenerateCode() {
--- a/src/toolbox.js
+++ b/src/toolbox.js
@@ -133,6 +133,12 @@ var Toolbox = GObject.registerClass({
     addTopic(id, title, iconName, widget) {
         const topic = new TopicButton({id, title, iconName});
         this._topicsList.add(topic);
+        if (GObject.Object.find_property.call(widget.constructor, 'needs-attention')) {
+            widget.bind_property('needs-attention',
+                topic, 'needs-attention', GObject.BindingFlags.SYNC_CREATE);
+        }
+
+
         widget.show_all();  // show_all() only propagates to current stack page
         this._topicsStack.add_titled(widget, id, title);
     }
--- a/src/topicButton.js
+++ b/src/topicButton.js
@@ -20,19 +20,33 @@ var TopicButton = GObject.registerClass({
         super._init(props);
 
         const overlay = new Gtk.Overlay();
-        const box = new Gtk.Box({
-            halign: Gtk.Align.FILL,
-            orientation: Gtk.Orientation.VERTICAL,
+        this._attentionSign = new Gtk.Revealer({
+            revealChild: false,
+            halign: Gtk.Align.END,
+            name: 'attention-sign',
+            valign: Gtk.Align.START,
         });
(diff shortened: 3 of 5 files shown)


MegFord:
I didn't see the icon shown in the mockups in Gtk. Should I request that from Design?

src/toolbox.js
```diff
@@ -212,6 +212,20 @@ var Toolbox = GObject.registerClass({
         this._infoLabel.label = text;
     }
 
+    setNeedsAttention(val, source) {
+        let topicRow = this._topicsList.get_selected_row();
```
ptomato: Instead of getting the selected row and special-casing it, if I am reading the code correctly, you have the ID of the topic in `source` so you could probably use `_findTopic(source)` to get the row and the child widget regardless of whether it is selected or not.

src/topicButton.js
```diff
@@ -32,7 +46,7 @@ var TopicButton = GObject.registerClass({
             hexpand: true,
             label: this._title,
         });
-
+        box.pack_start(this._attentionSign, false, false, 0);
```
ptomato: I might add the attention sign as an overlay child of `overlay` instead of a child of `box`. It looks like it'll be difficult to get it aligned as in the spec if it has to be within the GtkBox's padding.

src/topicButton.js
```diff
@@ -18,12 +18,26 @@ var TopicButton = GObject.registerClass({
...
     _init(props = {}) {
         super._init(props);
+        this._needsAttention = false;
 
         const overlay = new Gtk.Overlay();
         const box = new Gtk.Box({
             halign: Gtk.Align.FILL,
             orientation: Gtk.Orientation.VERTICAL,
         });
+        this._attentionSign = new Gtk.Revealer({
```
ptomato, line 23: This doesn't currently have a transition type set, are you waiting for a spec on that?

ptomato (changes requested):
Looks good, and I tested it out.

In addition to the comments on specific lines of code I have two general implementation suggestions that I think can make it easier to adapt to all the other toolboxes as well.

First, I would suggest making the `needs-attention` signal into a GObject property instead of a signal. A signal allows you to observe state changes, as you are doing here, but the property will allow that too (via the `notify::needs-attention` signal) as well as get/set the state directly. You could take the same property spec and getter/setter functions as used in `TopicButton`, and then instead of `_onNeedsAttention()` you could set the property.

That would also allow the second improvement that I'm thinking of, which would be to automatically bind the topic panel's needsAttention state to the topic button's needsAttention indicator state. In `Toolbox.addTopic()` you could put a `widget.bind_property('needs-attention', topic, 'needs-attention', ...)` and avoid having to connect it by hand for every topic.

This would require adding the property to all the other widgets passed to `addTopic()` in the other toolboxes, but you could also make it optional for the time being by checking if the `needs-attention` property exists before trying to bind it. Ultimately we do want the error UX improvements to apply to all the other toolboxes as well, so the more of it that you can implement in `src/toolbox.js` the better.

src/topicButton.js
MegFord: Yes. I left it as the default for that reason.

src/topicButton.js
MegFord: I tried that but overlay can only have one child. Should I try adding it as a child of the frame and see if i can get it to align correctly that way?

src/toolbox.js
MegFord: > Looks good, and I tested it out.
> 
> In addition to the comments on specific lines of code I have two general implementation suggestions that I think can make it easier to adapt to all the other toolboxes as well.
> 
> First, I would suggest making the `needs-attention` signal into a GObject property instead of a signal. A signal allows you to observe state changes, as you are doing here, but the property will allow that too (via the `notify::needs-attention` signal) as well as get/set the state directly. You could take the same property spec and getter/setter functions as used in `TopicButton`, and then instead of `_onNeedsAttention()` you could set the property.
> 
Are you suggesting that I add the same property that exists in the TopicButton class to the userFunction and controlPanel classes? The aspect of this that was difficult for me to wrap my head around was how I could access the property from the error checking code. I wasn't sure if I could access the property unless I had an instantiated TopicButton class in the userFunction/controlPanel classes. 
> That would also allow the second improvement that I'm thinking of, which would be to automatically bind the topic panel's needsAttention state to the topic button's needsAttention indicator state. In `Toolbox.addTopic()` you could put a `widget.bind_property('needs-attention', topic, 'needs-attention', ...)` and avoid having to connect it by hand for every topic.
> 
> This would require adding the property to all the other widgets passed to `addTopic()` in the other toolboxes, but you could also make it optional for the time being by checking if the `needs-attention` property exists before trying to bind it. Ultimately we do want the error UX improvements to apply to all the other toolboxes as well, so the more of it that you can implement in `src/toolbox.js` the better.

src/topicButton.js
ptomato: IIRC, overlays can have one main child but any number of overlay children. Overlay children are added with `add_overlay()` instead of `add()`.

src/toolbox.js
ptomato: Exactly, I was suggesting to add the same property to the UserFunction and ControlPanel classes.

It could be completely independent of the TopicButton instance. The only place where they would come into contact would be in `Toolbox.addTopic()` which would bind the TopicButton's property to the UserFunction or ControlPanel's property.

src/topicButton.js
```diff
@@ -66,13 +80,15 @@ var TopicButton = GObject.registerClass({
     }
 
     get needs_attention() {
-        return this._needsAttention;
+        return this._needs_attention;
     }
 
     set needs_attention(value) {
-        if ('_needsAttention' in this && this._needsAttention === value)
+        if (this._needs_attention === value)
```
ptomato: The casing change is fine, but I would still keep the `'_needs_attention' in this` check. IIRC otherwise you'll get a warning about accessing an undefined property the first time you set it.

src/LightSpeed/controlpanel.js
```diff
@@ -120,6 +125,7 @@ var LSCombinedTopic = GObject.registerClass({
             if (!(e instanceof SyntaxError || e instanceof ReferenceError))
                 throw e;
             this._variablesCodeview.setCompileResultsFromException(e);
+            this.set_property('needs-attention', true);
```
ptomato, line 128: TIL that `g_object_set_property()` sets the property _and_ emits the notify signal automatically! I will really have to rethink how I usually write GObject properties in GJS now.

I was going to suggest defining setter and getter functions like in the TopicButton class. You could still consider doing that, to make it so that the notify signal isn't emitted if the property already has the correct value, but on reflection I think it's probably fine as is.

ptomato (approved):
This is great, needs one small change. Feel free to merge it yourself when that's done.

src/topicButton.js
MegFord: Ah, I thought maybe there was some magic I didn't understand that made `this._$propertyName` refer to the GObject property directly. I'll change back the casing, thanks!

src/LightSpeed/controlpanel.js
MegFord: Yeah, I figured that we're doing something with the signal that's pretty inexpensive and we have the check in the set() for the topicButton. In general I agree that having the explicit methods is better.

PR #176, opened 2019-04-05
Lightspeed: Add spawnPowerup and activatePowerup topics

ptomato:
These are connected to the new exported global variables in Lightspeed.
They are defined per-level.

Names and icon assets for the topics are not added yet. These are using
placeholder names and assets.

https://phabricator.endlessm.com/T26025

### DIFF
--- a/src/LightSpeed/toolbox.js
+++ b/src/LightSpeed/toolbox.js
@@ -52,6 +52,18 @@ var LSToolbox = GObject.registerClass(class LSToolbox extends Toolbox {
             this._updateBeamTopic);
         this.hideTopic('updateBeam');
 
+        this._spawnPowerupTopic = new LSUserFunction('spawnPowerup');
+        this._spawnPowerupTopic.bindGlobal(this._global);
+        this.addTopic('spawnPowerup', 'NAME', 'go-home-symbolic',
+            this._spawnPowerupTopic);
+        this.hideTopic('spawnPowerup');
+
+        this._activatePowerupTopic = new LSUserFunction('activatePowerup');
+        this._activatePowerupTopic.bindGlobal(this._global);
+        this.addTopic('activatePowerup', 'NAME', 'weather-snow-symbolic',
+            this._activatePowerupTopic);
+        this.hideTopic('activatePowerup');
+
         this._updateLevelInfo();
         this.show_all();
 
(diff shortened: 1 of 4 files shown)


src/LightSpeed/toolbox.js
```diff
@@ -52,6 +52,18 @@ var LSToolbox = GObject.registerClass(class LSToolbox extends Toolbox {
             this._updateBeamTopic);
         this.hideTopic('updateBeam');
 
+        this._spawnPowerupTopic = new LSUserFunction('spawnPowerup');
+        this._spawnPowerupTopic.bindGlobal(this._global);
+        this.addTopic('spawnPowerup', 'NAME', 'go-home-symbolic',
+            this._spawnPowerupTopic);
+        // this.hideTopic('spawnPowerup');
```
danigm: Why is this line commented? It's for testing purpose?

danigm (changes requested):

manuq (changes requested):
As discussed, the `activatePowerup` changed to be global in the app side. So `getScope()` needs to be changed to be consistent with the rest of the user functions.

src/LightSpeed/toolbox.js
ptomato: Oh yeah, this should be uncommented now, glad you spotted it.

ptomato:
Comments addressed, and I also added a `pickOne()` function to the user scope as discussed on Slack.

src/LightSpeed/userFunction.js
```diff
@@ -38,6 +38,12 @@ const COMMON_SCOPE = {
         return min;
     },
 
+    pickOne(...choices) {
+        if (choices.length === 0)
+            throw new Error('You need at least one thing to pick from');
+        return choices[0];
```
manuq, line 44: This is always picking the first element. Shouldn't it pick a random element?

manuq (changes requested):

src/LightSpeed/userFunction.js
ptomato: It should in Lightspeed, but not in the toolbox! It should always be deterministic. See the comment above this part, on `random()`. Otherwise you might get an error, and then have it be "fixed" by adding a space somewhere :stuck_out_tongue:

(it would be better to actually check _all_ the choices each time, but that's quite out of scope)

manuq: Ah of course! I see now. /me is new to this part of the toolbox code.

manuq (approved):

PR #179, opened 2019-04-09
The flip from the Toolbox to Roster fails

xrgranados:
The flip from the Toolbox to Roster fails when the "corruption state" is not defined.

https://phabricator.endlessm.com/T25958

### DIFF
--- a/src/framework/model.js
+++ b/src/framework/model.js
@@ -277,13 +277,14 @@ class RaModelBase extends GObject.Object {
             const gameState = GameState.getDefault();
             const key = 'app.com_endlessm_Hackdex_chapter_one.corruption';
             let corruption;
+            let state;
             try {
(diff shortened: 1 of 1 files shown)


src/framework/model.js
```diff
@@ -279,11 +279,11 @@ class RaModelBase extends GObject.Object {
             let corruption;
             try {
                 corruption = (await gameState.Get(key)).deep_unpack();
+                const state = corruption.state.deep_unpack();
```
ptomato: This won't work, `state` will be declared locally to the `try` block and won't exist outside it. You'll need to declare it outside of the block and assign it inside the block.

ptomato (changes requested):

src/framework/model.js
xrgranados: Done.

PR #180, opened 2019-04-11
Lightspeed: Merge the two spawn topics into one (backup plan)

ptomato:
We now have two topics in the Lightspeed toolbox that have something to
do with spawning, and we want to merge them into one because having two
different ones is confusing.

This is the easier, backup-plan, version of what we actually want to do:
have the two functions in the same code view. For this backup plan, we
put both code views in the same topic.

https://phabricator.endlessm.com/T26244

### DIFF
--- a/src/LightSpeed/spawnTopic.js
+++ b/src/LightSpeed/spawnTopic.js
@@ -1,0 +1,34 @@
+/* exported LSSpawnTopic */
+
+const {GObject, Gtk} = imports.gi;
+
+const {LSUserFunction} = imports.LightSpeed.userFunction;
+
+var LSSpawnTopic = GObject.registerClass({
+    Properties: {
+        'needs-attention': GObject.ParamSpec.boolean('needs-attention', 'Needs attention',
+            'Display an indicator on the button that it needs attention',
+            GObject.ParamFlags.READWRITE, false),
+    },
+}, class LSSpawnTopic extends Gtk.Grid {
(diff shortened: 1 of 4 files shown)


cosimoc:
Merging this since the Clubhouse side already landed, and we don't want to leave things in a broken state for tomorrow.

MegFord:
@ptomato the needs-attention signal isn't working for this topic. It does work on master so it's a regression. 
Master:
master
Branch:
branch

src/LightSpeed/spawnTopic.js
```diff
@@ -0,0 +1,34 @@
+/* exported LSSpawnTopic */
+
+const {GObject, Gtk} = imports.gi;
+
+const {LSUserFunction} = imports.LightSpeed.userFunction;
+
+var LSSpawnTopic = GObject.registerClass({
+    Properties: {
```
MegFord, line 8: @ptomato Adding this here doesn't work because then you bind the signal on the TopicButton to it instead of to the property in the UserFunction that emits the 'needs-attention' signal. However, I don't know what the fix is here, since it seems like doing multiple bind_properties to one TopicButton doesn't work because the last call to bind_property takes precedence. I suspect you can figure out how to do it, though.

ptomato:
I opened a new ticket for the regression (https://phabricator.endlessm.com/T26291), since I didn't manage to fix it for April.

PR #181, opened 2019-04-15
Lockscreen: Manage GSS dbus service unknown error

danigm:
There's a race condition where we try to call a method for the
GameStateService just during the service is shuting down and this causes
some problems.

This patch manages this kind of errors in the getDictValueSync method
that is the one causing problems with the lock screen.

https://phabricator.endlessm.com/T25728

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -193,7 +193,7 @@ var Lockscreen = GObject.registerClass({
         this._activeSoundID = null;
     }
 
-    _updateLockStateWithLock() {
+    _updateLockStateWithLock(retries = null) {
         this._playbin.hasLock = !!this._lock;
 
         if (!this._lock)
@@ -201,8 +201,28 @@ var Lockscreen = GObject.registerClass({
 
         this._updateBackground();
 
-        if (!this._manager.isUnlocked(this._lock))
+        try {
+            if (!this._manager.isUnlocked(this._lock))
+                return;
+        } catch (e) {
+            const retry = retries === null ? 3 : retries;
+            const nextTryTimeout = 100;
+
+            if (retry > 0) {
+                logError(e, 'Error trying to load unlock state, ' +
+                    `trying again in ${nextTryTimeout} milisecond`);
+                GLib.timeout_add(GLib.PRIORITY_HIGH, nextTryTimeout,
+                    () => {
+                        this._updateLockStateWithLock(retry - 1);
+                        return GLib.SOURCE_REMOVE;
+                    });
+            } else {
+                logError(e, 'Error trying to load unlock state');
+            }
+
             return;
+        }
+
         this.locked = false;
     }
 
--- a/src/gameState.js
+++ b/src/gameState.js
@@ -68,6 +68,9 @@ var getDefault = (function() {
                     const dict = variant.deep_unpack();
                     return dict[property].deep_unpack();
                 } catch (e) {
+                    if (e.matches(Gio.DBusError, Gio.DBusError.SERVICE_UNKNOWN))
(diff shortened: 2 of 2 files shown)


xjuan:
This looks good to me, the only thing I would change is the timeout, I would use something less than 1 second, between 100 and 200 ms
That way it wont be noticeable for the user and it wont be flooding the computer either

src/lockscreen.js
```diff
@@ -201,8 +201,19 @@ var Lockscreen = GObject.registerClass({
...
+        try {
+            if (!this._manager.isUnlocked(this._lock))
+                return;
+        } catch (e) {
+            logError(e, 'Error trying to load unlock state, trying again in 100 milisecond');
+            GLib.timeout_add(GLib.PRIORITY_HIGH, 100,
+                () => {
+                    this._updateLockStateWithLock();
+                    return GLib.SOURCE_REMOVE;
+                });
```
joaquimrocha: This means this function will be tried until it succeeds. Should we limit the number of attempts? What happens if it's still trying and the user quits the application/toolbox?

src/lockscreen.js
```diff
@@ -201,8 +201,19 @@ var Lockscreen = GObject.registerClass({
 
         this._updateBackground();
 
-        if (!this._manager.isUnlocked(this._lock))
+        try {
+            if (!this._manager.isUnlocked(this._lock))
+                return;
+        } catch (e) {
+            logError(e, 'Error trying to load unlock state, trying again in 100 milisecond');
+            GLib.timeout_add(GLib.PRIORITY_HIGH, 100,
```
joaquimrocha: Nitpick: Don't repeat the number in the comment/timeout (as it's easy to get them unsynced in the future). Maybe the following works?
```
const next_try_timeout = 100;
logError(e, 'Error trying to load unlock state, trying again in %s milisecond', next_try_timeout);
GLib.timeout_add(GLib.PRIORITY_HIGH, next_try_timeout, ...)
```

joaquimrocha (commented):
The changes LGTM but see my comment regarding a hypothetical situation where the service is gone for good (unlikely but :woman_shrugging:  ).

src/lockscreen.js
danigm: Yes, you're right, if the service is broken or not installed for some reason this will be trying until the end of the app. I'll add a number of tries, like 3 for example.

src/lockscreen.js
danigm: Ok, let's do that.

MegFord:
This looks like the right approach to me. Merging.

src/lockscreen.js
```diff
@@ -193,16 +193,36 @@ var Lockscreen = GObject.registerClass({
...
             return;
 
         this._updateBackground();
 
-        if (!this._manager.isUnlocked(this._lock))
+        try {
+            if (!this._manager.isUnlocked(this._lock))
+                return;
+        } catch (e) {
+            const retry = retries === null ? 3 : retries;
```
joaquimrocha, line 208: This wouldn't be needed if the `retries` was `3` instead of `null` by default (and the code would be more readable too).

danigm: Yes, I didn't noticed that :man_facepalming:

danigm: I've just created a new PR to do that way:

https://github.com/endlessm/hack-toolbox-app/pull/184

joaquimrocha: > Yes, I didn't noticed that

No worried!

> I've just created a new PR to do that way:

Thank you! I'll take a look.

PR #183, opened 2019-04-16
Lightspeed: Discard changes on reset

ptomato:
If a code view contains an error, then it will not have been synced to
the Lightspeed game state. It's possible that if the "last good" state
was identical to the default code, then we don't get a notify signal
when the game resets its code to the default code. So, we need to
explicitly discard any discrepancies between the code view and the state
from Lightspeed.

https://phabricator.endlessm.com/T26290


MegFord (approved):
This fix is working for me

PR #184, opened 2019-04-17
Lockscreen: Use retries default value

danigm:
Instead of use null as default value and then check that to assign the
default number we can use directly the default number in the function
declaration and remove the check.

### DIFF
--- a/src/lockscreen.js
+++ b/src/lockscreen.js
@@ -193,7 +193,7 @@ var Lockscreen = GObject.registerClass({
         this._activeSoundID = null;
     }
 
-    _updateLockStateWithLock(retries = null) {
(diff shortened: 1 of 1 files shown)


joaquimrocha (approved):
Great! Thank you!

joaquimrocha:
Thank you for doing this PR @danigm . Since we are in freeze, I don't know if we should merge this. The changes seem simple, but if we've tested the previous code very thoroughly already, I'd prefer not to risk it.
Summoning @cosimoc for a decision.

cosimoc:
The original fix in https://phabricator.endlessm.com/T25728 was only merged yesterday, so I think this is a reasonable follow up. Let's get it in.

PR #185, opened 2019-04-17
Start RileyMaze toolbox

MegFord:


data/RileyMaze/panel.ui
```diff
@@ -0,0 +1,283 @@
...
+    <property name="can_focus">False</property>
+    <property name="border_width">20</property>
+    <property name="column_spacing">20</property>
+    <child>
+      <object class="GtkGrid" id="units">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="row_spacing">16</property>
+        <child>
+          <object class="Section">
```
ptomato, line 17: I would suggest giving the Section widgets a `border_width`. The spacing in this panel is a bit tight.

data/RileyMaze/icons/16x16/actions/powerup-symbolic.svg
```diff
@@ -0,0 +1,62 @@
...
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <sodipodi:namedview
```
ptomato, line 27: Minor comment and please consider this optional, but if you do happen to re-save the icons in Inkscape, you can save some space by saving them as "Plain SVG" which doesn't store all of Inkscape's editor state in them.

data/RileyMaze/panel.ui
```diff
@@ -0,0 +1,283 @@
...
+                </child>
+              </object>
+              <packing>
+                <property name="left_attach">0</property>
+                <property name="top_attach">1</property>
+                <property name="height">2</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkRadioButton" id="padAUp">
```
ptomato: As a heads up, we will probably need to retheme these radio buttons with something orange. (It was a guideline in the other toolboxes that all the clickable or draggable items should have some sort of orange indicator on them, but we didn't have any radio buttons in the other toolboxes as far as I recall.)

Check with Robin if she has any rough recommendations for this or leftover specs from a previous iteration. But it's fine to do this in a followup task, we should not wait on that for merging this.

data/RileyMaze/panel.ui
```diff
@@ -0,0 +1,283 @@
...
+        <property name="row_spacing">16</property>
+        <child>
+          <object class="Section">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="orientation">vertical</property>
+            <property name="heading">Pad A</property>
+            <property name="name">padADirection</property>
+            <child>
+              <object class="GtkMenuButton" id="padAAssetButton">
```
ptomato: It's probably OK to remove the Pad A and Pad B components. They look like they are pretty straightforward copies of Robot A and Robot B, so they should be pretty easy to recreate.

data/application.css
```diff
@@ -135,10 +135,6 @@ toolbox .left-column list row:active {
     margin-left: 12px;
 }
 
-.topic-revealer {
```
ptomato: This (and some other removals below) looks like it might be an error from rebasing?

src/RileyMaze/combinedTopic.js
```diff
@@ -0,0 +1,265 @@
...
+GObject.type_ensure(Section.$gtype);
+// Placeholders for the units
+const VALID_PAD_A = ['unicorn'];
+const VALID_PAD_B = ['unicorn'];
+const VALID_ROBOT_A = ['unicorn'];
+const VALID_ROBOT_B = ['unicorn'];
+const directions = {true: 'Up', false: 'Down'};
+const directionStr = {Up: true, Down: false};
+
+var RMZCombinedTopic = GObject.registerClass({
```
ptomato: I would name this `RMZUnitsTopic`. I can't remember why I named `LSCombinedTopic` that way, but that just goes to show it's not a good name to emulate :stuck_out_tongue_closed_eyes:

src/RileyMaze/combinedTopic.js
```diff
@@ -0,0 +1,265 @@
...
+        const robotBInfo = {};
+        VALID_ROBOT_B.forEach(id => {
+            robotBInfo[id] = `/com/endlessm/HackToolbox/RileyMaze/units/${id}.png`;
+        });
+        this._robotBAssetMenu = new PopupMenu(this._robotBAssetButton, robotBInfo,
+            Gtk.Image, 'resource');
+        this._variablesCodeview.connect('should-compile', this._compile.bind(this));
+    }
+
+    bindGlobal(model) {
```
ptomato: I would rename this to `bindGlobalModel`, to correspond with `unbindGlobalModel`. We intended to do that for Lightspeed and Fizzics as well, but since we introduced the split between global and per-level unbind methods as a critical last-minute bug fix for episode 2 (!) we did not want to risk renaming things that were not directly affected. In any case, we should use the new names going forward.

src/RileyMaze/combinedTopic.js
```diff
@@ -0,0 +1,265 @@
...
+            }
+            if (this._bindings) {
+                this._bindings.forEach(binding => binding.unbind());
+                this._bindings = null;
+            }
+            this._model = null;
+        }
+    }
+
+    _bindModel(model) {
```
ptomato: Similarly, this should be called `_bindLevelModel`.

src/RileyMaze/combinedTopic.js
```diff
@@ -0,0 +1,265 @@
...
+            if (dirBool === true)
+                this._robotBUp.set_active(true);
+            else if (dirBool === false)
+                this._robotBDown.set_active(true);
+        });
+        this._notifyHandler = model.connect('notify', this._onNotify.bind(this));
+        this._regenerateCode();
+    }
+
+    _on_active_changed(childName, button) {
```
ptomato: For consistency, `_onActiveChanged`.

src/RileyMaze/combinedTopic.js
```diff
@@ -0,0 +1,265 @@
...
+const SoundServer = imports.soundServer;
+
+GObject.type_ensure(Codeview.$gtype);
+GObject.type_ensure(Section.$gtype);
+// Placeholders for the units
+const VALID_PAD_A = ['unicorn'];
+const VALID_PAD_B = ['unicorn'];
+const VALID_ROBOT_A = ['unicorn'];
+const VALID_ROBOT_B = ['unicorn'];
+const directions = {true: 'Up', false: 'Down'};
```
ptomato: In the Riley Maze API document these were described as being `'up'` and `'down'` (without the capital letters.)

However, since that was written we've seen playtesters (and Hack employees) stumble over the quote chars for strings. We should check with Randy whether we should really add more, maybe we could make these into constants so you can just type `robotADirection = up;` in the code view without the quotes.

src/RileyMaze/model.js
```diff
@@ -0,0 +1,27 @@
+/* exported RMZModel */
+
+const {GObject} = imports.gi;
+const {ClippyWrapper} = imports.clippyWrapper;
+
+const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
+
+var RMZModel = GObject.registerClass({
+    Properties: {
+        scoreTarget: GObject.ParamSpec.double('scoreTarget', 'Score target', '',
```
ptomato: Score target can be deleted, I think.

src/RileyMaze/toolbox.js
```diff
@@ -0,0 +1,91 @@
...
+            new GLib.Variant('(sava{sv})', ['reset', [], {}]),
+            null, Gio.DBusCallFlags.NONE, -1, null);
+        this._unitTopic.reset();
+    }
+
+    _onGameStateChanged(self, senderName, [key, value]) {
+        if (!key.startsWith('rileymaze.topic.'))
+            return;
+        const topic = key.slice('rileymaze.topic.'.length);
+        // let {visible} = value.deep_unpack();
```
ptomato: This code is probably OK to be uncommented even though this game state isn't set yet.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,182 @@
...
+};
+
+const USER_FUNCTIONS = {
+    instructions: {
+        name: 'instructions',
+        args: [],
+        modelProp: 'instructionCode',
+        perLevel: true,
+        getScope() {
+            return Object.assign({}, COMMON_SCOPE);
```
ptomato: Here's where the `riley` object should go.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,182 @@
...
+        this._codeview.userFunction = userFunctionName;
+
+        this._codeview.connect('should-compile', this._compile.bind(this));
+
+        this.add(this._codeview);
+
+        this.get_style_context().add_class('user-function');
+    }
+
+    bindGlobal(model) {
```
ptomato: Same comments here about renaming `bindGlobal` and `_bindModel`.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,182 @@
...
+                message: `Expected a function named ${name}.`,
+                fixme: `function ${name}(${args.join('\n')}) {\n}\n`,
+            }]);
+            return;
+        }
+
+        try {
+            userFunction();
+        } catch (e) {
+            this._codeview.setCompileResultsFromException(e);
```
ptomato, line 239: In a few cases, the instructions codeview should work differently from all the other codeviews and call `_updateCode` with erroneous code, in order to pass it back to the game. This happens in cases where the game should display a glitched instruction tile. Double check with Randy when this is supposed to happen. My understanding is — typo in instruction name; too many instructions; too few instructions. But maybe he wants to have glitches in the game whenever there is an error at all, like in the case of invalid JS code?

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,182 @@
...
+        args: [],
+        modelProp: 'instructionCode',
+        perLevel: true,
+        getScope() {
+            return Object.assign({}, COMMON_SCOPE);
+        },
+    },
+};
+
+var RZUserFunction = GObject.registerClass({
```
ptomato: Should be RMZ for consistency

ptomato (changes requested):
This is working quite well so far, I had a lot of fun testing it together with the game branch. I think the biggest thing missing is validation of the `riley` object and its methods.

data/RileyMaze/icons/16x16/actions/powerup-symbolic.svg
MegFord: Yes, I'd forgotten about that, thanks! This is just a placeholder asset but I'll do that in the future when I have to edit in Inkscape :)

data/application.css
MegFord: I think this branch diverged from master a while ago and it's a conflict, yeah.

src/RileyMaze/combinedTopic.js
MegFord: Okay! I thought it was because it combined a codeview and widgets.

data/RileyMaze/panel.ui
ptomato: Border width didn't quite do what I was expecting, sorry for the wrong suggestion. I should have explained what I was aiming for in the first place instead of suggesting a solution.

In Lightspeed, there's a certain amount of space between the border of the Section widget and the widgets inside it:
Screenshot from 2019-04-23 17-53-43
In the robot Section, there are only a few pixels of space between the widgets and the border:
Screenshot from 2019-04-23 17-54-04
(and now that I look closer, it looks like the dark background behind "Robot A" doesn't extend all the way to the right side of the background box.)

I guess border width is not the right way to achieve that open spacing, but maybe take a look at Lightspeed to see what's different between these two widgets?

src/RileyMaze/RMZUnit.js
```diff
@@ -0,0 +1,219 @@
...
+
+    _compile() {
+        const code = this._variablesCodeview.text;
+
+        if (code === '')
+            return;
+
+        const scope = {
+            robotADirection: null,
+            robotBDirection: null,
```
ptomato, line 131: For `up` and `down` to work as variables instead of string constants, they will have to be defined in the scope object. Currently if I choose "up" or "down" in the UI and let the code be regenerated, then it looks like it works initially; but if I add a space or change something else so it recompiles, then I get an error: "up is not defined".

src/RileyMaze/RMZUnit.js
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    _searchForErrors(scope) {
+        const errors = [];
+        const PROPS_DEFAULTS = {
+            robotADirection: 'up',
+            robotBDirection: 'up',
+        };
+        Object.entries(PROPS_DEFAULTS).forEach(([prop, defaultValue]) => {
+            if (typeof scope[prop] !== 'string') {
```
ptomato: I'd suggest changing this to check whether the value is `'up'` or `'down'` and change the error message to say that the value must be up or down, since we don't allow just any string. (And we don't care if they use the string "up" or the variable "up")

src/RileyMaze/RMZUnit.js
```diff
@@ -0,0 +1,219 @@
...
+        if (this._updateSoundEnabled && !this._model.inReset &&
+            oldText !== this._variablesCodeview.text &&
+            timeMicrosec - this._lastCodeviewSoundMicrosec >= 100e3) {
+            SoundServer.getDefault().play('hack-toolbox/update-codeview');
+            this._lastCodeviewSoundMicrosec = timeMicrosec;
+        }
+    }
+
+    _regenerateCode() {
+        this._variablesCodeview.text = `
```
ptomato: Currently there's a newline on line 1 of this code block, if you put a `\` after the backquote it would remove the newline and be more consistent with other code views.

src/RileyMaze/RMZUnit.js
```diff
@@ -0,0 +1,219 @@
...
+robotADirection = ${this._model.robotADirection};
+robotBDirection = ${this._model.robotBDirection};
+`;
```
ptomato, line 222: Up to you if you want to change this as part of this PR or later, but you could consider adding the `robotAAsset` and `robotBAsset` variables here even if they're not hooked up to the game yet.

src/RileyMaze/model.js
```diff
@@ -0,0 +1,26 @@
...
+var RMZModel = GObject.registerClass({
+    Properties: {
+        robotADirection: GObject.ParamSpec.string('robotADirection', 'Robot A direction', '',
+            _propFlags, 'up'),
+        robotBDirection: GObject.ParamSpec.string('robotBDirection', 'Robot B direction', '',
+            _propFlags, 'up'),
+        instructionCode: GObject.ParamSpec.string('instructionCode', 'Instruction code', '',
+            _propFlags, ''),
+        levelCode: GObject.ParamSpec.string('levelCode', 'level code', '', _propFlags, ''),
+        rileyPosition: GObject.ParamSpec.string('rileyPosition', 'Riley position', '',
```
ptomato: I think `rileyPosition` and `goalPosition` could be passed along to the game as part of the `levelCode` string, since the player defines them in the code view. There isn't a separate UI element where they can be set, so if you keep them inside `levelCode` then you wouldn't have to do any syncing between their values in the player's code block and in this model.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+    height: 1080,
+    data: {},
+    riley: RILEY,
+};
+
+const LEVEL_EDIT_SCOPE = {
+    rileyPosition: 0,
+    goalposition: 2,
+    add(unit, x, y) {
+        if (x < 0 || x > 4)
```
ptomato: The maximum X is 7.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+            throw new RangeError(`${x} must be between 0 and 4`);
+        if (y < 0 || y > 4)
+            throw new RangeError(`${y} must be between 0 and 4`);
+        if (Number.isNaN(Number(x)))
```
ptomato, line 62: These checks should go before the previous checks, or they won't have an effect.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+            throw new TypeError(`${x} isn't a number`);
+        if (Number.isNaN(Number(x)))
+            throw new TypeError(`${x} isn't a number`);
+        if (typeof unit !== 'string')
```
ptomato: Here I'd check if the unit is not just any string, but one of the allowed units.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+const {Codeview} = imports.codeview;
+
+const FORWARD = 1;
+const UP = 2;
+const DOWN = 3;
+const JUMP = 4;
+const RILEY = {
+    // Keep this in sync with riley class in
+    // hack-toy-apps/com.endlessm.RileyMaze/parameters.js
+    queue: [],
```
ptomato, line 15: There was supposed to be an accompanying `ip` variable according to the API document, but I went ahead and just deleted that from the API document. It was secret anyway, and it's not actually needed.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+
+const {GObject, Gtk} = imports.gi;
+
+const {Codeview} = imports.codeview;
+
+const FORWARD = 1;
+const UP = 2;
+const DOWN = 3;
+const JUMP = 4;
+const RILEY = {
```
ptomato, line 12: According to the API document this should be a JS Proxy so that we can catch invalid property accesses and flag them as unknown instructions. But I think this is actually fine for now, as long as we do the JS Proxy on the game side so that we know when to add a glitched instruction to the game UI.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+    // Keep this in sync with riley class in
+    // hack-toy-apps/com.endlessm.RileyMaze/parameters.js
+    queue: [],
+    forward() {
+        this.queue.push(FORWARD);
```
ptomato, line 20: According to the API document, the instruction methods should throw an exception if the queue is already full, so we can display that to the player. (I saw that you are checking for >8 instructions below already, but throwing an exception here allows us to show the error dot at the right location.)

src/RileyMaze/userFunction.js
ptomato: Now that I think of it, we should probably put the `riley` object in its own scope (`INSTRUCTIONS_SCOPE`?) otherwise the Riley API will be available in the level editor too.

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        try {
+            userFunction();
+        } catch (e) {
+            this._codeview.setCompileResultsFromException(e);
+            return;
+        }
+        // FIXME: ASK FOR BETTER MESSAGE?
+        if (scope.riley.queue.length !== 8) {
```
ptomato: We have to move this somehow into a validation function that's separate for each user function panel. Otherwise the level editor will check if the Riley queue is full. (You can see this by adding a space or changing something else in the level code view so that it recompiles — you get "Instructions must have 8 moves")

ptomato (changes requested):
Nice work. I went back to the API document and updated a few things after testing this branch and after the discussion about string constants the other day.

Most of the comments are minor and this is pretty much mergeable as is. If you want to split off the other ones into follow up tickets, so that we can get it merged tomorrow, that's fine with me.

ptomato:
Oh, I almost forgot — before merging this should use the new name `com.endlessm.Sidetrack` instead of `com.endlessm.RileyMaze`

src/RileyMaze/userFunction.js
```diff
@@ -0,0 +1,247 @@
...
+    add(unit, x, y) {
+        if (Number.isNaN(Number(x)))
+            throw new TypeError(`${x} isn't a number`);
+        if (Number.isNaN(Number(y)))
+            throw new TypeError(`${y} isn't a number`);
+        if (!UNITS.includes(unit))
+            throw new TypeError(`${unit} isn't a valid unit`);
+        if (x < 0 || x > 7)
+            throw new RangeError(`${x} must be between 0 and 7`);
+        if (y < 0 || y > 7)
```
ptomato: The maximum X is now correct, but the maximum Y should be 4.

ptomato (approved):
As we discussed on Slack, I found only one thing that really should be addressed before merging. It's a minor fix, go ahead and merge this after fixing it. I'll leave a list of other things in the Phabricator ticket and you can decide what to do with them.

PR #186, opened 2019-04-18
Lightspeed: show topics when game state is unreachable

ptomato:
Emergency fix. Topics default to shown, when the game state service is
unreachable. This may sometimes show them when they are not yet supposed
to be shown, but at least they will not be hidden when the player needs
them in order to progress.

https://phabricator.endlessm.com/T25728


joaquimrocha (approved):
LGTM.

PR #187, opened 2019-04-23
Lightspeed: Fix needs-attention forwarding in Spawn topic

ptomato:
Previously, I forgot to make the spawn topic's needs-attention property
depend on the needs-attention properties of both of the code views that
it contains.

https://phabricator.endlessm.com/T26291


MegFord (approved):
Nice solution! So simple.

PR #188, opened 2019-04-23
framework: Fix missing defaults

ptomato:
Various fixes found while developing the toolbox for Leviathan.

https://phabricator.endlessm.com/T25507


src/framework/defaults.js
```diff
@@ -21,6 +21,18 @@ const KEYS = {
 const COLORS = Symbol('colors');
 
 const DEFAULTS = {
+    unknown: {
```
manuq, line 29: Why not use 'com.endlessm.dinosaurs.en' as default? Just a suggestion.

src/framework/model.js
```diff
@@ -403,7 +403,7 @@ function ensureModelClass(appId, defaults) {
         return _createdClasses.get(appId);
 
     const ModelClass = GObject.registerClass({
-        GTypeName: `RaModel_for_${appId.replace(/./g, '_')}`,
+        GTypeName: `RaModel_for_${appId.replace(/\./g, '_')}`,
```
manuq, line 417: :+1:

manuq (approved):

src/framework/defaults.js
ptomato: I figured it would be better to have the default be equal to the default colors in the modular framework default theme. The choice of "encyclopedia" as the default logo is pretty arbitrary, though.

PR #189, opened 2019-04-26
Sidetrack: Fix Error handling and UI

MegFord:
Fixes outstanding issues in Sidetrack Toolbox:
Checking for <8 instructions
Showing instructions error on right line
Validation of the level editor code
Fix Topic title shadow

https://phabricator.endlessm.com/T26109

### DIFF
--- a/src/RileyMaze/userFunction.js
+++ b/src/RileyMaze/userFunction.js
@@ -8,42 +8,57 @@ const FORWARD = 1;
 const UP = 2;
 const DOWN = 3;
 const JUMP = 4;
+const MAX_QUEUE_LEN = 8;
+
+class InstructionError extends Error {
+}
 
 const RILEY = {
     // Keep this in sync with riley class in
     // hack-toy-apps/com.endlessm.RileyMaze/parameters.js
     queue: [],
     forward() {
-        if (this.queue.length >= 8)
-            throw new SyntaxError('Instructions must have 8 moves');
-
         this.queue.push(FORWARD);
+
+        if (this.queue.length > MAX_QUEUE_LEN)
+            throw new InstructionError('Instructions must have 8 moves.');
     },
 
     up() {
-        if (this.queue.length >= 8)
-            throw new SyntaxError('Instructions must have 8 moves');
-
         this.queue.push(UP);
+
+        if (this.queue.length > MAX_QUEUE_LEN)
+            throw new InstructionError('Instructions must have 8 moves.');
     },
 
     down() {
-        if (this.queue.length >= 8)
-            throw new SyntaxError('Instructions must have 8 moves');
-
         this.queue.push(DOWN);
+
+        if (this.queue.length > MAX_QUEUE_LEN)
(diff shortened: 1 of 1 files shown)


src/RileyMaze/userFunction.js
```diff
@@ -198,7 +186,10 @@ ${code}
         const {name, args, getScope} = USER_FUNCTIONS[this._codeview.userFunction];
         const code = this._codeview.text;
         const scope = getScope();
-        scope.riley.queue.length = 0;
+
+        if (scope.hasOwnProperty('riley'))
```
ptomato: This seems like it should be moved into the Instructions topic's `getScope` function, since the `riley` object only applies there. I'd like to avoid special-casing things in the UserFunction class because I think the common user function code should eventually be factored out of Lightspeed and Sidetrack.

src/RileyMaze/userFunction.js
```diff
@@ -240,6 +232,22 @@ ${code}
             return;
         }
 
+        if (scope.hasOwnProperty('riley') && scope.riley.queue.length !== MAX_QUEUE_LEN) {
```
ptomato: Likewise I'd prefer to avoid this special-case as well, if possible, because `if (scope.hasOwnProperty('riley'))` is basically saying `if (this is the "instructions" user function)`. Maybe it would work to add a `validate()` function to the descriptions in `USER_FUNCTIONS` similar to `getScope()`, and put this code there?

src/RileyMaze/userFunction.js
```diff
@@ -240,6 +232,22 @@ ${code}
...
+            // Add lines for the function name
+            let lineNum = MAX_QUEUE_LEN + 2;
```
ptomato: I don't feel very comfortable with guessing the line number, as it can easily lead to unintentionally confusing error messages when the player does something we don't expect:
Screenshot from 2019-04-26 12-22-46
If I'm reading the code correctly, you removed the error-throwing code from the methods of the Riley object because of https://phabricator.endlessm.com/T26436 ? I think I would prefer here to reinstate those, and instead fix the underlying bug in a separate PR. Throwing the error seems to be the correct way to get the error message at the correct line, but it just happens to not work at the moment...

src/RileyMaze/userFunction.js
```diff
@@ -240,6 +232,22 @@ ${code}
...
+            if (scope.riley.queue.length < MAX_QUEUE_LEN)
+                lineNum = scope.riley.queue.length + 1;
```
ptomato: IIRC in the discussion on Randy's spec, we landed on showing the error on the last line of the code view in the case of too few instructions.

ptomato (changes requested):
The UI widgets in the units panel look much better now!

src/RileyMaze/userFunction.js
MegFord: Yeah, I agree, thanks!

ptomato (approved):
Great, thanks!

PR #192, opened 2019-04-30
Add lockscreen to topic buttons

MegFord:
https://phabricator.endlessm.com/T26409


src/toolbox.js
```diff
@@ -130,19 +130,28 @@ var Toolbox = GObject.registerClass({
...
         }
 
-
         widget.show_all();  // show_all() only propagates to current stack page
         this._topicsStack.add_titled(widget, id, title);
     }
 
+    addTopicKeys(id, appName, lockNum) {
+        const [topicRow, topic] = this._findTopic(id); // eslint-disable-line
+        topic.topicLock.key = `item.key.${appName}.${lockNum}`;
```
ptomato: I would suggest passing in the key name and lock name to this function, and moving the logic of figuring out what the names should be to the caller. The reason for this is because we need to support keys that will open more than one lock.

src/RileyMaze/toolbox.js
```diff
@@ -67,21 +71,6 @@ var RMZToolbox = GObject.registerClass(class RMZToolbox extends Toolbox {
         this._unitTopic.reset();
     }
 
-    _onGameStateChanged(self, senderName, [key, value]) {
```
ptomato, line 70: I don't see any callers of this method being removed elsewhere in this commit, so I suspect it should either be reinstated or the callers should be removed.

src/RileyMaze/toolbox.js
```diff
@@ -47,9 +51,9 @@ var RMZToolbox = GObject.registerClass(class RMZToolbox extends Toolbox {
 
     bindWindow(win) {
         win.get_style_context().add_class('RileyMaze');
-        this._unitTopic.bindWindow(win);
```
ptomato, line 50: It looks like `RMZUnit.bindWindow()` is obsolete, so you should go ahead and remove that method as well. Additionally it seems like it might be better to do that in a separate commit since at first glance it looks like it's fixing a separate bug.

src/RileyMaze/toolbox.js
```diff
@@ -92,6 +81,8 @@ var RMZToolbox = GObject.registerClass(class RMZToolbox extends Toolbox {
         }
 
         this._instructTopic.unbindGlobalModel();
+        this._unitTopic.unbindGlobalModel();
```
ptomato, line 79: Looks like adding these calls is fixing a separate bug? If so, put it in a separate commit.

src/toolbox.js
```diff
@@ -130,19 +130,28 @@ var Toolbox = GObject.registerClass({
...
                 topic, 'needs-attention', GObject.BindingFlags.SYNC_CREATE);
         }
 
-
         widget.show_all();  // show_all() only propagates to current stack page
         this._topicsStack.add_titled(widget, id, title);
     }
 
+    addTopicKeys(id, appName, lockNum) {
+        const [topicRow, topic] = this._findTopic(id); // eslint-disable-line
```
ptomato: In general, prefer to disable only specific warnings with eslint, e.g. `// eslint-disable-line no-restricted-syntax`. That said, if I remove this comment it looks like the warning is that `topicRow` is unused; you can fix that here with `const [, topic] = this._findTopic(id);` and below with `const [topicRow] = this._findTopic(id);` (you don't have to name all the parameters that you are unpacking from an array, unlike Python)

src/toolbox.js
```diff
@@ -161,6 +170,15 @@ var Toolbox = GObject.registerClass({
         topic.get_style_context().remove_class('reveal');
     }
 
+    selectTopic(id) {
+        const [topicRow, topic] = this._findTopic(id); // eslint-disable-line
+        // Set milisecs in timeout to length of  lockscreen video
```
ptomato: There appears to be a `done` signal on the playbin widget which it looks like you could connect to instead of hardcoding the video length. I'm not sure exactly how that works, but Juan Pablo would have more details if you need them.

Nice touch, by the way, to select the topic once the lockscreen opens.

src/topicButton.js
```diff
@@ -18,8 +22,14 @@ var TopicButton = GObject.registerClass({
 }, class TopicButton extends Gtk.Frame {
     _init(props = {}) {
         super._init(props);
-
         const overlay = new Gtk.Overlay();
+
+        if (this._lockscreen) {
+            this.topicLock = new Lockscreen({visible: true});
+            overlay.add_overlay(this.topicLock);
```
ptomato, line 29: I was surprised that the lockscreen widgets work when added as an overlay, instead of having the thing they are locking added as their child widget! But I think this is indeed the best way to do it. Otherwise you have to deal with widgets having different parents depending on whether there is a lockscreen or not.

ptomato (changes requested):
A few minor comments. On the whole this looks good and works well in testing!

src/toolbox.js
MegFord: I did try removing the unused parameters (that is how I coded it originally) but it didn't work. I'll try again.

src/RileyMaze/toolbox.js
MegFord: I hadn't added any callers because the game state wasn't in place for the game yet. Since I'm locking the topic buttons instead of hiding them and the quest isn't using 'topic' in any keys I didn't end up using the method for anything.

ptomato: Oh, okay.

ptomato (approved):

ptomato:
Good to go, thanks!

PR #193, opened 2019-05-01
Remaining bits of Leviathan hackability

ptomato:
https://phabricator.endlessm.com/T25507


MegFord (approved):
The css is loading very slowly for me when the text is encrypted, but it looks to me like the regex, etc. is being done in the fastest way possible so I don't have any complaints.

PR #201, opened 2019-05-05
Allow Disabling of topics

MegFord:
https://phabricator.endlessm.com/T26408

### DIFF
--- a/data/application.css
+++ b/data/application.css
@@ -161,7 +161,7 @@ toolbox .left-column list row:active {
     padding: 4px 17px;
 }
 
-row:hover .topic {
+row:hover .topic:not(.insensitive) {
     background-color: rgba(32, 46, 65, 0.8);  /* lighten rgba(6, 20, 39, 0.8) 10% */
 }
 
@@ -176,12 +176,33 @@ row:hover .topic {
     padding: 4px;
 }
 
-row:selected .topic {
+row:selected .topic:not(.insensitive) {
     background-color: #f18c22;
 }
 
-row:selected:hover .topic {
+row:selected:hover .topic:not(.insensitive) {
     background-color: #ffa63c;  /* lighten #f18c22 10% */
+}
+
+.topic.insensitive {
+    background-color:  rgba(0, 0, 0, 0.3);
+    border-radius: 25px;
+    min-height: 154px;
+    min-width: 154px;
+    outline-width: 0;
+    transition: background-color 200ms;
+}
+
+.topic.insensitive label {
+    background-color: rgba(0, 0, 0, 0.1);
+    color: rgba(109, 110, 112, 1.0);
+    border-radius: 12px;
+    margin-top: 12px;
+    padding: 4px;
+}
+
+.topic.insensitive image {
+    color: rgba(109, 110, 112, 0.4);
 }
 
 .sourceview {
--- a/data/application.css
+++ b/data/application.css
@@ -185,20 +185,12 @@ row:selected:hover .topic:not(.insensitive) {
 }
 
 .topic.insensitive {
-    background-color:  rgba(0, 0, 0, 0.3);
-    border-radius: 25px;
-    min-height: 154px;
-    min-width: 154px;
-    outline-width: 0;
-    transition: background-color 200ms;
+    background-color: rgba(0, 0, 0, 0.3);
 }
 
 .topic.insensitive label {
     background-color: rgba(0, 0, 0, 0.1);
     color: rgba(109, 110, 112, 1.0);
-    border-radius: 12px;
-    margin-top: 12px;
-    padding: 4px;
 }
 
 .topic.insensitive image {
--- a/src/Fizzics/toolbox.js
+++ b/src/Fizzics/toolbox.js
@@ -14,6 +14,7 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox 
         this._controlPanel.bindModel(this._model);
         this.addTopic('main', 'Tools', 'preferences-other-symbolic', this._controlPanel);
         this.show_all();
+        this.selectTopic('main');
 
         this.connect('reset', () => {
             Gio.DBus.session.call_sync(
--- a/src/toolbox.js
+++ b/src/toolbox.js
@@ -24,7 +24,7 @@ var Toolbox = GObject.registerClass({
 }, class Toolbox extends Gtk.Grid {
     _init(appId, props = {}) {
         super._init(props);
-
+        this.appId = appId;
         const leftColumn = new Gtk.Grid({
             orientation: Gtk.Orientation.VERTICAL,
             valign: Gtk.Align.FILL,
@@ -70,7 +70,7 @@ var Toolbox = GObject.registerClass({
         });
         this._topicsList = new Gtk.ListBox({
             halign: Gtk.Align.START,
-            selectionMode: Gtk.SelectionMode.BROWSE,
+            selectionMode: Gtk.SelectionMode.SINGLE,
         });
         this._topicsList.connect('row-selected', this._onRowSelected.bind(this));
         topicsScroll.add(this._topicsList);
@@ -130,14 +130,23 @@ var Toolbox = GObject.registerClass({
         this.setBusy(false);
     }
 
-    addTopic(id, title, iconName, widget, lockscreen = false) {
-        const topic = new TopicButton({id, title, iconName, lockscreen});
+    addTopic(id, title, iconName, widget, lockscreen = false, sensitive = true) {
+        const topic = new TopicButton({id, title, iconName, lockscreen, sensitive});
         this._topicsList.add(topic);
 
         if (GObject.Object.find_property.call(widget.constructor, 'needs-attention')) {
             widget.bind_property('needs-attention',
                 topic, 'needs-attention', GObject.BindingFlags.SYNC_CREATE);
         }
+        topic.connect('notify::revealed', () => {
+            this._notifyRevealed(topic, id);
+        });
+
+        topic.connect('notify::sensitive', () => {
+            this._notifySensitive(topic, id);
+        });
+
+        topic.dbusRegister(this.appId);
 
         widget.show_all();  // show_all() only propagates to current stack page
         this._topicsStack.add_titled(widget, id, title);
@@ -152,6 +161,15 @@ var Toolbox = GObject.registerClass({
         });
     }
 
+    _notifyRevealed(topic, id) {
+        if (topic.revealed) {
+            this.revealTopic(id);
+            this.selectTopic(id);
+        } else {
+            this.hideTopic(id);
+        }
+    }
+
     _findTopic(id) {
         const rows = this._topicsList.get_children();
         const topicRow = rows.find(row => row.get_child().id === id);
@@ -163,7 +181,7 @@ var Toolbox = GObject.registerClass({
         const [topicRow, topic] = this._findTopic(id);
 
         if (this._topicsList.get_selected_row() === topicRow)
-            this._topicsList.select_row(null);
+            this._topicsList.unselect_row(topicRow);
 
         topicRow.noShowAll = true;
         topicRow.hide();
@@ -205,10 +223,13 @@ var Toolbox = GObject.registerClass({
         }
 
         const topic = row.get_child();
-        this._headerLabel.label = topic.title;
-        this._headerImage.iconName = topic.icon_name;
-        this._topicsStack.visibleChildName = topic.id;
-        this._setMinimized(false);
+        if (topic.sensitive) {
+            this._headerLabel.label = topic.title;
+            this._headerImage.iconName = topic.icon_name;
+            this._topicsStack.visibleChildName = topic.id;
+            this._setMinimized(false);
+        }
+        topic.emit('clicked');
     }
 
     _onResetClicked() {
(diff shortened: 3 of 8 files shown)


src/topicButton.js
```diff
@@ -1,9 +1,22 @@
 /* exported TopicButton */
-
-const {GObject, Gtk} = imports.gi;
+const {GObject, Gtk, GLib, Gio} = imports.gi;
 
 const {Lockscreen} = imports.lockscreen;
 
+const TopicIface = `
+<node>
+  <interface name="com.endlessm.Toolbox.topic">
```
manuq: Early-tester review: I think the interface name should be: com.endlessm.**Hack**Toolbox.topic (to match the app name).

manuq (changes requested):

src/topicButton.js
MegFord: Done, thanks!

src/Fizzics/toolbox.js
```diff
@@ -14,6 +14,7 @@ var FizzicsToolbox = GObject.registerClass(class FizzicsToolbox extends Toolbox
         this._controlPanel.bindModel(this._model);
         this.addTopic('main', 'Tools', 'preferences-other-symbolic', this._controlPanel);
         this.show_all();
+        this.selectTopic('main');
```
ptomato, line 17: Great, I'm glad you were able to fix this bug in all toolboxes.

I was wondering if it could be made harder to forget in the future by selecting the topic in `addTopic` if it was the first one added, but I think this is fine for the time being as it is getting a bit off-topic for this PR.

src/topicButton.js
```diff
@@ -109,4 +133,65 @@ var TopicButton = GObject.registerClass({
...
+
+    set sensitive(value) {
+        if ('_sensitive' in this && this._sensitive === value)
+            return;
+
+        this._sensitive = value;
+        this.notify('sensitive');
+
+        if ('_sensitive' in this && !this._sensitive) {
+            this.get_style_context().remove_class('topic');
```
ptomato: I would keep the `.topic` style class always on, and add/remove an `.insensitive` style class when the property changes. That way you can also avoid duplicating some of the CSS rules.

src/topicButton.js
```diff
@@ -109,4 +133,65 @@ var TopicButton = GObject.registerClass({
...
+    }
+
+    set sensitive(value) {
+        if ('_sensitive' in this && this._sensitive === value)
+            return;
+
+        this._sensitive = value;
+        this.notify('sensitive');
+
+        if ('_sensitive' in this && !this._sensitive) {
```
ptomato: The first half of the check is not necessary because you set `this._sensitive` a few lines above.

src/topicButton.js
```diff
@@ -1,9 +1,22 @@
...
+const TopicIface = `
+<node>
+  <interface name="com.endlessm.HackToolbox.topic">
+    <method name="reveal">
+      <arg type="b" direction="in" name="revealed"/>
+      <arg type="b" direction="out" name="revealed"/>
+    </method>
+    <property name="sensitive" type="b" access="readwrite"/>
+    <signal name="clicked">
+    <arg type="s" name="button"/>
```
ptomato: Nitpick: Indentation is off here.

src/topicButton.js
```diff
@@ -1,9 +1,22 @@
...
+const {GObject, Gtk, GLib, Gio} = imports.gi;
 
 const {Lockscreen} = imports.lockscreen;
 
+const TopicIface = `
+<node>
+  <interface name="com.endlessm.HackToolbox.topic">
+    <method name="reveal">
+      <arg type="b" direction="in" name="revealed"/>
+      <arg type="b" direction="out" name="revealed"/>
```
ptomato: Is the out argument necessary? From reading the code I don't see a case where it is going to be different than the in argument.

src/toolbox.js
```diff
@@ -205,10 +221,13 @@ var Toolbox = GObject.registerClass({
...
-        this._headerLabel.label = topic.title;
-        this._headerImage.iconName = topic.icon_name;
-        this._topicsStack.visibleChildName = topic.id;
-        this._setMinimized(false);
+        if (topic.sensitive) {
+            this._headerLabel.label = topic.title;
+            this._headerImage.iconName = topic.icon_name;
+            this._topicsStack.visibleChildName = topic.id;
+            this._setMinimized(false);
+            topic.emit('clicked');
```
ptomato: I think we will want to emit the `clicked` signal no matter whether the topic is sensitive or not. I'm not 100% sure but I think Ben wanted to be able to trigger dialogue when the user clicked on a disallowed topic.

src/toolbox.js
```diff
@@ -237,8 +256,19 @@ var Toolbox = GObject.registerClass({
         this._infoLabel.label = text;
     }
 
+    _notifySensitive(topic, id) {
+        const child = this._topicsStack.get_child_by_name(id);
+        if (topic.sensitive)
```
ptomato: This is probably one of those things that would have been clearer if the GJS documentation was working ... but I think `child.sensitive = topic.sensitive` will work.

src/topicButton.js
```diff
@@ -18,10 +31,20 @@ var TopicButton = GObject.registerClass({
         'needs-attention': GObject.ParamSpec.boolean('needs-attention', 'Needs attention',
             'Display an indicator on the button that it needs attention',
             GObject.ParamFlags.READWRITE, false),
+        sensitive: GObject.ParamSpec.boolean('sensitive', 'sensitive',
```
ptomato, line 33: I was wondering if this would interfere with the `sensitive` property of `GtkWidget`. I guess not, since this branch works fine when I test it?

src/toolbox.js
```diff
@@ -237,8 +256,19 @@ var Toolbox = GObject.registerClass({
...
+        else
+            child.set_sensitive(false);
+    }
+
     // Can be overridden by subclasses to do any work on window close
     shutdown() {
-        void this;
+        this._topicsList.get_children().forEach(row => {
+            const topic = row.get_child();
+            topic._dbus.unexport();
```
ptomato: I would add a method to TopicButton to do this, in order to avoid accessing its private `_dbus` property.

src/toolbox.js
ptomato: Or now that I think about it, I guess what Robin asked for this morning is that clicking on an insensitive topic still works the same as when the topic is sensitive, but just with the panel widget insensitive? Otherwise you can click away from the topic but you can't click back to it.

This is especially apparent with one-topic toolboxes like Fizzics. If you disable the one topic, then collapse the panel with the minimize arrow, then you can never expand the panel again.

src/topicButton.js
```diff
@@ -109,4 +133,65 @@ var TopicButton = GObject.registerClass({
...
+            this.get_style_context().add_class('topic');
+        }
+    }
+
+    _dbusRegister() {
+        const objPath = Gio.Application.get_default().get_dbus_object_path();
+        const keys = Object.keys(Gio.Application.get_default()._windows);
+        const winName = keys[keys.length - 1].split('.');
+        const connId = winName[winName.length - 1];
+        this.objectPath = `${objPath}/window/${connId}/topic/${this._id}`;
```
ptomato: On hindsight, I think I gave you bad advice about this when I suggested to match it to the toolbox window object path. I think instead of `connId` here we should have the full app ID (although dots may need to be replaced by underscores to make it a valid object path.) Otherwise the quest has to guess or hardcode what the `connId` is for each app.

Later we should go back and change the object paths of the hack toolbox window objects, for this same reason, but that's outside the scope of this issue in any case.

Accessing `keys[keys.length - 1]` is a bit risky IMO anyway because a newer window could have been created in the meantime. I think you can get the app ID like this: `this.get_toplevel().target_app_id` (see `src/window.js`) but only after the topic button has been added to the widget hierarchy, so you might want to make `dbusRegister` public and call it from `addTopic`. (and indeed I'm not sure how I got there, but I ended up in a state where I had Sidetrack topics exported as `/com/endlessm/HackTopics/window/Fizzics/...`)

src/topicButton.js
```diff
@@ -1,9 +1,22 @@
 /* exported TopicButton */
-
-const {GObject, Gtk} = imports.gi;
+const {GObject, Gtk, GLib, Gio} = imports.gi;
 
 const {Lockscreen} = imports.lockscreen;
 
+const TopicIface = `
+<node>
+  <interface name="com.endlessm.HackToolbox.topic">
```
ptomato: This is bikeshedding, but I noticed in d-feet that most other DBus interface names were capitalized as in `com.example.AppName.SomeObjectName`, so I'd suggest capitalizing `Topic` for consistency.

ptomato (changes requested):
I think this is the right idea and it seems like it's certainly already enough to unblock Manuel and Ben if they want to test against this branch. Mostly minor comments. The two things that I think need attention are:
- Interaction between collapsing the panel and sensitivity, we may need to check back with Robin on this one
- Naming of the DBus object path, I got into a weird state where the objects were named wrong (see comment for more info)

src/topicButton.js
manuq: I haven't seen if [`GtkWidget.set_sensitive()`](https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-set-sensitive) used in this PR. But if that's the case, the `:disabled` CSS class should be added automatically:
```
.topic:disabled label {
  background-color: rgba(0, 0, 0, 0.1);
}
```

src/topicButton.js
MegFord: Yeah the only time it would be different is if it somehow failed to set the property.

src/toolbox.js
MegFord: My understanding was that we should 1) not show the topic panel when the insensitive button is clicked; and 2) when we have a topic panel that is already shown (for example by default or because the topic was previously selected by the user) then the child widgets should all be disabled

MegFord: I see your point though, and that isn't a case I thought of during my meeting with Robin. I guess we can either implement it the way you mentioned or we could also disable the panel collapse button. I'll check with Robin.

manuq: Yes, I can confirm, Ben designed the quest to display messages when the player clicks on insensitive topics. At least from the quest, the clicked signal is only relevant for disabled topics.

```
        # if the user flips the app and clicks on a coding panel, Faber responds as follows:
        # LIGHTSPEEDKEY_PANEL_LOCKED - if the panel they've clicked on is not available this level
        # LIGHTSPEEDKEY_PANEL_GAME- if they click on the Game panel
        # LIGHTSPEEDKEY_PANEL_SPAWN - if they click on the Spawn panel
        # LIGHTSPEEDKEY_PANEL_ASTEROID - if they click on the Asteroid panel
        # LIGHTSPEEDKEY_PANEL_SPINNER - if they click on the Spinner panel
        # LIGHTSPEEDKEY_PANEL_SQUID - if they click on the Squid panel
        # LIGHTSPEEDKEY_PANEL_BEAM - if they click on the Beam panel
        # LIGHTSPEEDKEY_PANEL_POWERUPS - if they click on the PowerUps panel
```

manuq: Sorry, there are messages for clicking in insensitive and sensitive topics. Like:
> This panel isn't available on this level!  How can you solve it with another panel?
> The code in these panels creates behaviors of an enemy type. This panel works for Squids

MegFord: Thanks for the clarification!

src/topicButton.js
ptomato: I think in that case the method should throw an exception.

src/topicButton.js
ptomato: It's not entirely clear until Meg and Robin sync again, but I suspect that "insensitive" in the case of  the topic button will need to mean "_looks_ insensitive but actually still reacts to clicks" in which case we can't use that.

MegFord: I need to be able to send the click signal, yes.

src/topicButton.js
```diff
@@ -109,4 +131,63 @@ var TopicButton = GObject.registerClass({
...
+        this._sensitive = value;
+        this.notify('sensitive');
+
+        if (!this._sensitive)
+            this.get_style_context().add_class('insensitive');
+        else if (this._sensitive)
+            this.get_style_context().remove_class('insensitive');
+    }
+
+    dbusRegister(appId) {
```
ptomato, line 164: Great, this is a much better way to get the app ID into this function than I thought there was going to be.

data/application.css
```diff
@@ -176,14 +176,35 @@ row:hover .topic {
...
 }
 
-row:selected:hover .topic {
+row:selected:hover .topic:not(.insensitive) {
     background-color: #ffa63c;  /* lighten #f18c22 10% */
 }
 
+.topic.insensitive {
+    background-color:  rgba(0, 0, 0, 0.3);
+    border-radius: 25px;
```
ptomato: These lines that apply to both sensitive and insensitive topics can be removed, as `.topic.insensitive` inherits and overrides rules from `.topic`.

ptomato (approved):
Great, I was no longer able to confuse the toolbox about which topics belonged to which app.

I have a minor comment about deduplicating some of the CSS. Since I didn't get around to reviewing this before the end of your day, and we discussed getting it into the playthrough tomorrow, I'll just make and test that change myself, then merge it.

PR #202, opened 2019-05-06
Hackdex 2: Implement new hover spec

xrgranados:
https://phabricator.endlessm.com/T26460

PR #203, opened 2019-05-06
Hackdex 2: meow sounds for Leviathan

xrgranados:
https://phabricator.endlessm.com/T26493


ptomato:
Same question as in the hackdex PR, please separate this one from #202.

PR #208, opened 2019-05-09
Insert "< Roster" on footer.

xrgranados:
https://phabricator.endlessm.com/T25042

### DIFF
--- a/src/framework/gen.js
+++ b/src/framework/gen.js
@@ -561,6 +561,83 @@ $body-font: ${model.font};
     .LayoutScrolling--maincontainer {
         background-color: $background-medium-color;
     }
+
+    .Nav-LayoutBox {
+        min-width: 1905px;
+        min-height: 415px;
+    }
+
+    .Gtk-button-back {
+        color: $accent-light-color;
+        font-family: $title-font;
+        font-size: 25px;
+        margin-left: 95px;
+        margin-top: 65px;
+        min-height: 150px;
+        min-width: 150px;
+    }
+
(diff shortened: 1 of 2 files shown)


src/framework/gen.js
```diff
@@ -561,6 +561,83 @@ $body-font: ${model.font};
...
+                    color: $primary-light-color;
+                }
+
+                frame {
+                    background-position: center -377px;
+                    border: 7px solid white;
+                }
+            }
+
+            &.CardHackdex-0:hover .CardHackdex__title {
```
ptomato, line 620: I would like to avoid setting the background for these cards in the CSS, because if we change the sorting order of the cards in the hacking panel then we'll get the wrong backgrounds, as the names will be sorted differently but the backgrounds will be tied to the physical position of the card on screen. The content metadata already provides a "thumbnail URI" property for each card, can we use that?

ptomato (changes requested):
This is looking a lot better, but I would like to see if we can use the content metadata for the cards' backgrounds instead of putting them in CSS. Otherwise, it will become more difficult to hack the app effectively.

src/framework/gen.js
xrgranados: @ptomato I have it clear now and agree on avoiding using CSS for this cards. The metadata already provides the "thumbnail URI" property for each card, we actually can use that.  I`ll be working on it during the week and deliver as soon as possible.

danigm:
Closing old PR

PR #211, opened 2019-05-15
Sidetrack & LightSpeed: Scope notify signal per model for code update

MegFord:
https://phabricator.endlessm.com/T26141

### DIFF
--- a/src/LightSpeed/controlpanel.js
+++ b/src/LightSpeed/controlpanel.js
@@ -28,7 +28,6 @@ var LSCombinedTopic = GObject.registerClass({
 }, class LSCombinedTopic extends Gtk.Grid {
     _init(props = {}) {
         this._lastCodeviewSoundMicrosec = 0;
-
         super._init(props);
 
         const shipInfo = {};
@@ -80,10 +79,8 @@ var LSCombinedTopic = GObject.registerClass({
 
     _bindModel(model) {
         this._unbindLevelModel();
-
         this._model = model;
         const flags = GObject.BindingFlags.BIDIRECTIONAL | GObject.BindingFlags.SYNC_CREATE;
-
         const bindingInfo = {
             scoreTarget: this._scoreTargetAdjustment,
             astronautSize: this._astronautSizeAdjustment,
@@ -93,8 +90,8 @@ var LSCombinedTopic = GObject.registerClass({
         };
         this._bindings = Object.entries(bindingInfo).map(([prop, target]) =>
             model.bind_property(prop, target, 'value', flags));
-
-        this._notifyHandler = model.connect('notify', this._onNotify.bind(this));
+        this._notifyHandler = this._model.connect(
+            'notify', (object, pspec) => this._onNotify(pspec));
         this._regenerateCode();
     }
 
@@ -142,7 +139,7 @@ var LSCombinedTopic = GObject.registerClass({
         this._variablesCodeview.setCompileResults([]);
         this.set_property('needs-attention', false);
 
-        // Block the normal notify handler that updates the code view, since we
+        // Block the normal notify handlers that update the code view, since we
         // are propagating updates from the codeview to the GUI. Instead,
         // connect a temporary handler that lets us know if anything actually
         // did change.
@@ -199,8 +196,16 @@ var LSCombinedTopic = GObject.registerClass({
         return errors;
     }
 
-    _onNotify() {
-        this.reset();
+    _onNotify(pspec) {
+        const props = [
+            'scoreTarget',
+            'astronautSize',
+            'shipAsset',
+            'shipSize',
+            'shipSpeed',
+        ];
+        if (props.includes(pspec.name))
+            this.reset();
     }
 
     reset() {
--- a/src/RileyMaze/RMZUnit.js
+++ b/src/RileyMaze/RMZUnit.js
@@ -69,6 +69,7 @@ var RMZUnitsTopic = GObject.registerClass({
                 this._model.disconnect(this._notifyHandler);
                 this._notifyHandler = null;
             }
+
             if (this._bindings) {
                 this._bindings.forEach(binding => binding.unbind());
                 this._bindings = null;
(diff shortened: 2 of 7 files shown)


MegFord:
@ptomato one thing I'm a little worried about is that adding/blocking/removing so many signals in LightSpeed will slow things down. Can you let me know if you see any slowdown, and if so, if you can recommend a better way to re-write the notify signals in that app?

src/LightSpeed/controlpanel.js
```diff
@@ -28,7 +28,7 @@ var LSCombinedTopic = GObject.registerClass({
 }, class LSCombinedTopic extends Gtk.Grid {
     _init(props = {}) {
         this._lastCodeviewSoundMicrosec = 0;
-
+        this._notifyHandler = {};
```
ptomato: I think it would be better to use `new Map()` here instead of `{}`, since you will get a more idiomatic way to tell if the map is empty than "length of keys is 0 and constructor is Object", as well as setting/deleting keys.

src/LightSpeed/controlpanel.js
```diff
@@ -62,14 +62,14 @@ var LSCombinedTopic = GObject.registerClass({
...
 
     _unbindLevelModel() {
         if (this._model) {
-            if (this._notifyHandler) {
-                this._model.disconnect(this._notifyHandler);
-                this._notifyHandler = null;
-            }
+            if (Object.keys(this._notifyHandler).length === 0 &&
+                this._notifyHandler.constructor === Object)
+                this._disconnectNotify();
```
ptomato: I'm not sure, but this condition seems backwards on first reading? It seems like disconnectNotify should be called only if notifyHandler is not empty. The code works, though, so maybe I'm just not understanding it. A comment might be useful in that case :smile:

src/LightSpeed/controlpanel.js
```diff
@@ -80,24 +80,47 @@ var LSCombinedTopic = GObject.registerClass({
 
     _bindModel(model) {
         this._unbindLevelModel();
-
         this._model = model;
         const flags = GObject.BindingFlags.BIDIRECTIONAL | GObject.BindingFlags.SYNC_CREATE;
-
-        const bindingInfo = {
+        this.bindingInfo = {
```
ptomato: It's unlikely, but just to be sure I'd move this initializer into the constructor, in case any of the other methods that use this were called before this method.

Alternatively, it could end up being easier to iterate through `this._notifyHandler` in those other methods, instead of through `this.bindingInfo`, since you only need the keys. If you take the suggestion above about using a map, `Map` has a `forEach()` method which you could use.

src/LightSpeed/controlpanel.js
```diff
@@ -62,14 +62,14 @@ var LSCombinedTopic = GObject.registerClass({
             return;
         this._currentLevel = this._global.currentLevel;
         this._bindModel(this._global.getModel(this._currentLevel));
+        this.set_property('needs-attention', false);
```
ptomato: Is this related to the fix or just a general improvement? If the latter, split it out into a separate commit.

src/RileyMaze/RMZUnit.js
```diff
@@ -83,7 +90,7 @@ var RMZUnitsTopic = GObject.registerClass({
         this._model = model;
         this._robotAUp.connect('toggled',
             this._onActiveChanged.bind(this, 'robotADirection'));
-        this._model.connect('notify::robotADirection', () => {
+        this.robotANotify = this._model.connect('notify::robotADirection', () => {
```
ptomato: Prefix these property names with an underscore to indicate they are private. Optionally, you could use a notifyHandlers map like in Lightspeed to be more future-proof.

ptomato (changes requested):
OK, it took me a while and a lot of poking around to understand this, but I now see why the changes to the variables codeviews are necessary as well as fixing the user functions codeviews. That was the part I didn't understand the other day when we talked on Slack about this. Thanks for the thorough investigation, I found this bug really difficult to wrap my head around.

ptomato:
PS. I didn't notice a slowdown, but I also didn't test it on the Hack laptop.

If it slows down, one way to solve it differently could be to continue connecting to the full `notify` signal in the variables codeviews, and just check the property name at the start of the signal handler, and return early if it's not a property we're interested in. I'd still use the `notify::propName` in the user functions codeviews though, since that really is only one property.

src/LightSpeed/controlpanel.js
MegFord: Ha, yeah, I forgot to ! this check :/

src/LightSpeed/controlpanel.js
MegFord: I'm actually seeing the issue this fixed in a few places. I'l break it out into another bug.

MegFord: So I see that I also fixed this in RMZUnits in the PR so I'll leave it in if that's okay. This line fixes a regression caused by using the specific notify signals. In the past the generic notify was being called and that called reset() which changed this property to false when the level advanced.

src/LightSpeed/controlpanel.js
```diff
@@ -199,8 +197,17 @@ var LSCombinedTopic = GObject.registerClass({
         return errors;
     }
 
-    _onNotify() {
-        this.reset();
+    _onNotify(pspec) {
+        const props = {
```
ptomato: A more idiomatic way to do this would be
```js
const props = ['scoreTarget', 'astronautSize', 'shipAsset', 'shipSize', 'shipSpeed'];
if (props.includes(pspec.name))
    this.reset();
```

src/RileyMaze/RMZUnit.js
```diff
@@ -191,8 +192,14 @@ var RMZUnitsTopic = GObject.registerClass({
         return errors;
     }
 
-    _onNotify() {
-        this.reset();
+    _onNotify(obj, pspec) {
+        const props = {
```
ptomato: Ditto here.

ptomato (approved):
Works fine, go ahead and merge it after making the change, if you agree with it.

src/LightSpeed/controlpanel.js
MegFord: Oh, nice, I didn't realize JS had a way to check for an object in an array that wasn't O(n). I guess it was added since I used JS regularly.

ptomato: Oh, good point, I assume this is still O(n)... but with 5 entries I wouldn't be too concerned about it.

If O(1) is really necessary I'd suggest using a Set:
```js
const props = new Set(['scoreTarget', blah, blah...]);
if (props.has(pspec.name)) ...
```

MegFord: Yeah in this case it makes no difference. I just got used to using O(1) when I was doing operations on large data sets, so it was force of habit. Thanks for the info!

PR #212, opened 2019-05-15
Cipher text preview

ptomato:
https://phabricator.endlessm.com/T26655


MegFord:
Looks good! This design is much clearer.

PR #217, opened 2019-05-17
Sidetrack: Add Robot Images to Show Up/Down State

MegFord:
https://phabricator.endlessm.com/T26660


ptomato (changes requested):
This seems to work correctly, but the flipped robot assets don't match the assets in Sidetrack. So these assets might not be the latest ones?

Screenshot from 2019-05-17 12-10-26

(This is with both robots flipped. note the red forehead lights are visible on the robots in the game, but not in the toolbox asset, and the blue forehead lights are visible on the toolbox asset, but not in the game.)

MegFord:
Here's a screenshot with the new images from Anita robots

ptomato (approved):
Assets looking good now, thanks!

PR #218, opened 2019-05-17
Restart Game State Service on Rebuild

MegFord:
Joaquim mentioned this morning that we should put this in our script since GSS only shuts down after 5 minutes now instead of 10 seconds.


ptomato (approved):
:+1:

PR #220, opened 2019-05-21
Sidetrack: No sound effects if code error is set initially

MegFord:
https://phabricator.endlessm.com/T26687

### DIFF
--- a/src/RileyMaze/toolbox.js
+++ b/src/RileyMaze/toolbox.js
@@ -66,6 +66,7 @@ var RMZToolbox = GObject.registerClass(class RMZToolbox extends Toolbox {
             new GLib.Variant('(sava{sv})', ['reset', [], {}]),
             null, Gio.DBusCallFlags.NONE, -1, null);
         this._unitTopic.reset();
+        this._levelTopic.discardChanges();
     }
 
     shutdown() {
--- a/src/RileyMaze/userFunction.js
+++ b/src/RileyMaze/userFunction.js
@@ -145,8 +145,8 @@ var RMZUserFunction = GObject.registerClass({
 
         this._codeview = new Codeview();
         this._codeview.userFunction = userFunctionName;
-
-        this._codeview.connect('should-compile', this._compile.bind(this));
+        this._codeview.connect('should-compile',
+            (widget, userInitiated) => this._compile(userInitiated));
 
         this.add(this._codeview);
 
@@ -201,7 +201,6 @@ var RMZUserFunction = GObject.registerClass({
         const {modelProp} = USER_FUNCTIONS[this._codeview.userFunction];
         this._notifyHandler = model.connect(`notify::${modelProp}`, this._setCode.bind(this));
         this._setCode();
-        this._compile();
     }
 
     _setCode() {
@@ -238,7 +237,7 @@ ${code}
         }
     }
 
-    _compile() {
+    _compile(userInitiated = true) {
         const {name, args, getScope, validate} = USER_FUNCTIONS[this._codeview.userFunction];
         const code = this._codeview.text;
         const scope = getScope();
@@ -257,7 +256,7 @@ ${code}
             this.set_property('needs-attention', true);
             if (!(e instanceof SyntaxError || e instanceof ReferenceError))
                 throw e;
-            this._codeview.setCompileResultsFromException(e);
+            this._codeview.setCompileResultsFromException(e, userInitiated);
             return;
         }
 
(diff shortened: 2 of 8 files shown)


src/RileyMaze/userFunction.js
```diff
@@ -238,7 +238,17 @@ ${code}
         }
     }
 
-    _compile() {
+    _compile(isUserInitiated = true) {
+        // Sometimes this is called by the signal
```
ptomato: I think a good improvement would be to disambiguate this, so that it's never in question whether the first argument is the codeview or the boolean. You could change the line where the signal is connected, to work like this:
```js
this._codeview.connect('should-compile', () => this._compile(...));
```
However it's an open question what to put in the place of `...`. I think the least convoluted way might be to add a boolean parameter to the `should-compile` signal so that you can do
```js
this._codeview.connect('should-compile',
    (widget, userInitiated) => this._compile(userInitiated));
```
and then in codeview.js you can change `this.compile` to take that parameter as well, so that you can call it with `true` from `_onBufferChanged()` and with `false` from `set text()`.

On the other hand, that might require changing all the consumers of `should-compile` in all the variables codeviews in other toolboxes, so maybe we should save it for a future refactor. Although it's possible that it might be as easy as adding `if (!isUserInitiated) return;` to the `should-compile` handlers in all the variables codeviews.

src/codeview.js
```diff
@@ -224,7 +224,7 @@ var Codeview = GObject.registerClass({
                     this._buffer.text = value;
             });
             this._cached_ast = null;
-            this.setCompileResults([]);
+            this.setCompileResults([], false);
```
ptomato: I think the assumption under which I originally wrote this line (that programmatically-set text will always have blank compile results) no longer holds. I think it would be good to instead call `this.compile()` here (and that means you cuold remove `this._compile()` from the constructor of `UserFunction`), but that would require adding the boolean flag as a parameter to the `should-compile` signal as suggested above.

src/codeview.js
```diff
@@ -453,29 +453,32 @@ var Codeview = GObject.registerClass({
...
+            // Only play sounds when user compiles the code
+            if (userInitiated) {
+                if (this._numErrors)
+                    SoundServer.getDefault().play('codeview/code-error-appearance');
+                else
+                    SoundServer.getDefault().play('codeview/code-error-disappearance');
+            }
         }
 
         this._recordErrorMetrics(results)
```
ptomato: I think we should probably only record metrics if the error is user initiated, as well.

src/RileyMaze/userFunction.js
```diff
@@ -201,7 +201,7 @@ var RMZUserFunction = GObject.registerClass({
         const {modelProp} = USER_FUNCTIONS[this._codeview.userFunction];
         this._notifyHandler = model.connect(`notify::${modelProp}`, this._setCode.bind(this));
         this._setCode();
-        this._compile();
+        this._compile(false);
```
ptomato: This might be able to be removed altogether, see previous comment.

ptomato (changes requested):
This is looking really good. I originally thought you might have to add some internal state such as `this._lastCompileWasUserInitiated` but this design, where the flag is not stored but passed around from the place where the action is initiated, is much cleaner.

I have a couple of suggestions that I think might streamline things even more, although it might bring in some extra work in which case maybe we should save it for a future refactor. Up to you.

src/RileyMaze/userFunction.js
MegFord: Thanks, this is an excellent improvement!

src/RileyMaze/toolbox.js
```diff
@@ -66,6 +66,7 @@ var RMZToolbox = GObject.registerClass(class RMZToolbox extends Toolbox {
             new GLib.Variant('(sava{sv})', ['reset', [], {}]),
             null, Gio.DBusCallFlags.NONE, -1, null);
         this._unitTopic.reset();
+        this._levelTopic.discardChanges();
```
ptomato, line 69: Does this need an accompanying `this._instructionsTopic.discardChanges()`?

ptomato (approved):
One question that I'm not sure about, otherwise feel free to merge it at your convenience. I'll go ahead and test it manually now, and let you know if I find anything. Thanks!

src/RileyMaze/toolbox.js
MegFord: Yeah I didn't have time to get to the bottom of this, but the instructionsTopic was already working with the reset button before this change so adding it here called it twice and added empty lines to the codeview.

PR #222, opened 2019-05-31
Auto indentation

ptomato:
https://phabricator.endlessm.com/T26421

### DIFF
--- a/src/codeview.js
+++ b/src/codeview.js
@@ -527,7 +527,7 @@ var Codeview = GObject.registerClass({
         const {start} = funcDecl.body.loc;
         // Workaround for mozjs52 bug; funcDecl.body.loc.end is incorrect
         const {end} = funcDecl.loc;
-        const lines = this._buffer.text.split('\n');
+        const lines = this._buffer.text.split(/\r\n|[\n\v\f\r\x85\u2028\u2029]/);
 
         // Add 2 to the start column, 1 because slice() is inclusive, and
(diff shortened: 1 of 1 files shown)


src/codeview.js
```diff
@@ -527,7 +527,7 @@ var Codeview = GObject.registerClass({
         const {start} = funcDecl.body.loc;
         // Workaround for mozjs52 bug; funcDecl.body.loc.end is incorrect
         const {end} = funcDecl.loc;
-        const lines = this._buffer.text.split('\n');
+        const lines = this._buffer.text.split(/\r\n|[\n\v\f\r\x85\u2028\u2029]/);
```
MegFord, line 649: Why did you choose to split here rather than replace with a unix line ending and then split? I see from your comment that the standard is to handle all the possibilities, as you do here, but in reality does having a windows line ending in the code cause issues for the gjs interpreter or phaser or is it handled fine?

ptomato: GJS seems to deal as expected with alternative line endings such as `\r\n` or `\u2028`, counting them correctly in its line count. The idea here was to count lines in exactly the same way as GJS does, since we're relying on its line count to pick out the lines of the function body.

Replacing all line endings with `\n` as they are inserted, would also work.

I have a slight preference for doing it this way, since I hope we can remove this hacky code entirely in https://phabricator.endlessm.com/T26818 — but I could go either way.

PR #223, opened 2019-06-03
Sidetrack, LightSpeed: Add post-condition validation

MegFord:
https://phabricator.endlessm.com/T26819

### DIFF
--- a/src/LightSpeed/userFunction.js
+++ b/src/LightSpeed/userFunction.js
@@ -60,23 +60,78 @@ const COMMON_SCOPE = {
     },
 };
 
-const COMMON_SPAWN_SCOPE = {
-    // Keep this in sync with SpawnScope class in
-    // hack-toy-apps/com.endlessm.LightSpeed/userScope.js
-    ticksSinceSpawn: 0,
-};
+class Point {
+    constructor() {
+        this.x = 0;
+        this.y = 0;
+    }
+}
+// Set method that checks type for number properties of objects
+function propNumFactory(obj, propName) {
+    Object.defineProperty(obj, propName, {
+        set(newVal) {
+            if (Number.isNaN(Number(newVal)))
+                throw new TypeError(`${propName} must be a number.`);
+            else if (Number(newVal) < 0)
+                throw new RangeError(`${propName} must be a positive number.`);
+            obj[`_${propName}`] = newVal; // eslint-disable-line no-param-reassign
+        },
+        get() {
+            return obj[`_${propName}`];
+        },
+    });
+}
+
+const playPos = new Point();
+propNumFactory(playPos, 'x');
+propNumFactory(playPos, 'y');
+
+const enemyPos = new Point();
+propNumFactory(enemyPos, 'x');
+propNumFactory(enemyPos, 'y');
+
+const enemyVelocity = new Point();
+propNumFactory(enemyVelocity, 'x');
+propNumFactory(enemyVelocity, 'y');
 
 const COMMON_UPDATE_SCOPE = {
     // Keep this in sync with UpdateEnemyScope class in
     // hack-toy-apps/com.endlessm.LightSpeed/userScope.js
     playerShip: {
-        position: {x: 0, y: 0},
+        position: playPos,
     },
     enemy: {
-        position: {x: 0, y: 0},
-        velocity: {x: 0, y: 0},
+        position: enemyPos,
+        velocity: enemyVelocity,
         data: {},
     },
+};
+
+const shipPos = new Point();
+propNumFactory(shipPos, 'x');
+propNumFactory(shipPos, 'y');
+
+const POWERUP = {
+    ship: {
+        position: shipPos,
+        invulnerableTimer: 0,
+        shrinkTimer: 0,
+        attractTimer: 0,
+    },
+    powerUpType: 0,
+    blowup: 'blowup',
+    invulnerable: 'invulnerable',
+    upgrade: 'upgrade',
+    shrink: 'shrink',
+    attraction: 'attraction',
+    engine: 'engine',
+    blowUpEnemies() {},  // eslint-disable-line no-empty-function
+};
+
+var COMMON_SPAWN_SCOPE = {
+    // Keep this in sync with SpawnScope class in
+    // hack-toy-apps/com.endlessm.LightSpeed/userScope.js
+    ticksSinceSpawn: 0,
 };
 
 const USER_FUNCTIONS = {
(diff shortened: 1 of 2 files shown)


src/LightSpeed/userFunction.js
```diff
@@ -301,10 +384,17 @@ ${code}
             return;
         }
 
+        try {
```
ptomato: I'd suggest folding this into the previous stanza, as far as I can see we may as well validate the function the first time we execute it.

Or, alternatively, move the execution of the function outside of `validate()` and call something like `validate(scope, ret)` instead.

src/LightSpeed/userFunction.js
```diff
@@ -53,33 +53,100 @@ const COMMON_SCOPE = {
...
         getScope() {
-            return Object.assign({}, COMMON_SPAWN_SCOPE, COMMON_SCOPE);
+            const ENEMY_SCOPE = Object.assign({}, COMMON_SPAWN_SCOPE, COMMON_SCOPE);
+            propNumFactory(ENEMY_SCOPE, 'ticksSinceSpawn', ticksSinceSpawn);
+            return ENEMY_SCOPE;
+        },
+        validate(func) {
+            const ret = func();
+            // Don't throw an error when returning null
+            if (ret && !COMMON_SCOPE.enemyTypes.includes(ret))
```
ptomato: I'd suggest making this check explicitly `ret === null`. If they delete the return statement altogether, then they'll return `undefined` and my guess is we should throw the `TypeError` in that case.

src/LightSpeed/userFunction.js
```diff
@@ -53,33 +53,100 @@ const COMMON_SCOPE = {
     },
 };
 
-const COMMON_SPAWN_SCOPE = {
-    // Keep this in sync with SpawnScope class in
-    // hack-toy-apps/com.endlessm.LightSpeed/userScope.js
-    ticksSinceSpawn: 0,
+// Set method that checks type for number properties of objects
+function propNumFactory(obj, propName, valName) {
```
ptomato: I'm finding it difficult to understand what `valName` is for... it seems to me the value should be stored somewhere on the object, or does this already do that somehow?

src/LightSpeed/userFunction.js
```diff
@@ -53,33 +53,100 @@ const COMMON_SCOPE = {
...
+
+const NUMTYPE_HANDLER = {
+    set(obj, prop, value) {
+        if (prop in obj) {
+            if (Number.isNaN(Number(value)))
+                throw new TypeError(`Value for ${prop} must be a number.`);
+            else
+                return Reflect.set(obj, prop, value);
+        } else {
+            throw new TypeError(`Unknown property ${prop}.`);
```
ptomato: I'm hesitating about whether we do want to throw here or silently allow it as a normal Javascript object would. My understanding from reading the code is that this will mark on the third line of the following script as an error, is that correct?
```js
enemy.position.x = 3;
enemy.position.y = 4;
enemy.position.z = 5;  // !
```

I think it's probably a net positive, and it's OK to keep it for now and see how it turns out, although this seems like something that Chaim might feel strongly one way or the other on.

src/LightSpeed/userFunction.js
```diff
@@ -53,33 +53,100 @@ const COMMON_SCOPE = {
...
+        if (name in target)
+            return Reflect.get(target, name, receiver);
+
+        throw new TypeError(`Unknown property ${name}`);
+    },
 };
 
+// Proxy for position values
+const POSITION = {x: 0, y: 0};
+const POS_PROXY = new Proxy(POSITION, NUMTYPE_HANDLER);
```
ptomato: :+1: your idea of using proxies and/or getters/setters for these where possible, instead of checking the values at the end of the script. It makes it possible to mark the errors as coming from the correct line. I didn't think of that.

src/LightSpeed/userFunction.js
```diff
@@ -53,33 +53,100 @@ const COMMON_SCOPE = {
...
+
 const COMMON_UPDATE_SCOPE = {
     // Keep this in sync with UpdateEnemyScope class in
     // hack-toy-apps/com.endlessm.LightSpeed/userScope.js
     playerShip: {
-        position: {x: 0, y: 0},
+        position: POS_PROXY,
```
ptomato: When I test it, it doesn't seem to be a problem, but from reading the code I would expect all these `POS_PROXY` properties to store their x and y values in the same target object, `POSITION`. Out of curiosity, do you know why that's not happening?

src/RileyMaze/userFunction.js
```diff
@@ -77,6 +80,19 @@ const COMMON_SCOPE = {
     data: {},
 };
 
+const POSITION_HANDLER = {
+    set(obj, prop, value) {
+        if ((prop === 'rileyPosition' || prop === 'goalPosition') &&
+            (value < 0 || value > 4))
+            throw new RangeError(`${prop} must be between 0 and 4.`);
+        else if ((prop === 'rileyPosition' || prop === 'goalPosition') &&
```
ptomato: I'd suggest moving the NaN check to be first, before checking the range.

ptomato (changes requested):
Looks good, only minor comments.

I think we may want to think about designing something more systematic for the future, where scope objects are generated out of some sort of spec. I could imagine it getting tedious to write these proxies/getters/setters repeatedly. But for our purposes at the moment, this is great, and I think will really contribute towards getting players unstuck.

src/LightSpeed/userFunction.js
MegFord: Yes, that's correct. I added it because I thought it was in keeping with the types of errors we give for the scope properties in Sidetrack. I consider this part of the code to be potentially hard for kids to follow so I thought it would be better to opt for feedback on meeting the lesson goals rather than flexibility in this case. If it turns out to more confusing then I'm totally fine with removing it.

src/codeview.js
```diff
@@ -166,11 +166,15 @@ var Codeview = GObject.registerClass({
         this._buffer.tagTable.add(this._errorStyle);
 
         this._view = new GtkSource.View({
+            autoIndent: true,
```
ptomato: Seems like a rebase error, this is from a different branch?

src/LightSpeed/userFunction.js
ptomato: I'm still having trouble puzzling out how `valName` works ... maybe a comment would help?

Optional, I think you could also use this to define the `x` and `y` properties on `Point` to make it less verbose.

ptomato (changes requested):

src/LightSpeed/userFunction.js
ptomato: Huh, I see what you mean in the other instance of this check where you have to allow `undefined`, since we have an empty `handlePowerup` function in many Lightspeed levels ... we'd either have to change all the defaults to `return null` or something like that. Given that, I think what you have is fine for the time being.

ptomato (approved):
Thanks!

PR #224, opened 2019-06-07
LightSpeed, Sidetrack: Refactor UserFunction classes

MegFord:
https://phabricator.endlessm.com/T26820

### DIFF
--- a/src/userFunction.js
+++ b/src/userFunction.js
@@ -1,0 +1,196 @@
+/* exported UserFunction */
+
+const {GObject, Gtk} = imports.gi;
+
+const {Codeview} = imports.codeview;
+
+class InstructionError extends Error {
+}
+
+var UserFunction = GObject.registerClass({
+    Properties: {
+        'needs-attention': GObject.ParamSpec.boolean('needs-attention', 'Needs attention',
+            'Display an indicator on the button that it needs attention',
+            GObject.ParamFlags.READWRITE, false),
+    },
+}, class UserFunction extends Gtk.Frame {
+    _init(userFunctionName, props = {}) {
+        super._init(props);
+        this._codeview = new Codeview();
+        this._codeview.userFunction = userFunctionName;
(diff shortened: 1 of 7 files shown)


src/userFunction.js
```diff
@@ -0,0 +1,196 @@
+/* exported UserFunction */
+
+const {GObject, Gtk} = imports.gi;
+
+const {Codeview} = imports.codeview;
+
+class InstructionError extends Error {
```
ptomato, line 7: I think we should maybe rename this to `NonFatalError` so that it's not specific to Sidetrack, and do something like `class InstructionError extends UserFunction.NonFatalError {}` in the Sidetrack toolbox.

src/userFunction.js
```diff
@@ -0,0 +1,196 @@
...
+            (widget, userInitiated) => this._compile(userInitiated));
+
+        this.add(this._codeview);
+
+        this.get_style_context().add_class('user-function');
+    }
+
+    bindGlobalModel(model) {
+        this._global = model;
+        const {perLevel, modelProp} = this.USER_FUNCTIONS[this._codeview.userFunction];
```
ptomato, line 31: I'm not sure we should still keep the `USER_FUNCTIONS` dict after this refactor, or at least maybe we should keep it out of the userFunction.js file. Its data properties such as `name` and `args` could all be passed into the constructor, and `getScope()` and `validate()` could be methods of `UserFunction` that get subclassed and overridden.

The `perLevel` part I'm a bit hesitant about encoding the concept of "levels" and a `currentLevel` parameter into `UserFunction`. On the other hand, it is repeated code that ought to be deduplicated to reduce mistakes. Maybe it could be put into a subclass, so that we can create either a `UserFunction` which makes no assumptions about levels, or a `PerLevelUserFunction` which does make those assumptions?

ptomato (changes requested):

danigm:
Closing old PR

PR #226, opened 2019-06-14
Sketchbook prototype

ptomato:
Here's what I had for the Sketchbook app prototype. Not sure if it even makes sense to merge it, but I thought I should at least put it in a branch.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -30,7 +30,7 @@ function _toolboxClassForAppId(targetAppId) {
         return imports.OperatingSystemApp.toolbox.OSToolbox;
     case 'com.endlessm.Sidetrack':
         return imports.RileyMaze.toolbox.RMZToolbox;
-    case 'name.ptomato.AbsentRhythm':
+    case 'com.endlessm.Sketchbook':
         return imports.sketchbook.toolbox.SketchToolbox;
     default:
         return imports.hacktoolbox.hacktoolbox.DefaultHackToolbox;
--- a/src/app.js
+++ b/src/app.js
@@ -30,6 +30,8 @@ function _toolboxClassForAppId(targetAppId) {
         return imports.OperatingSystemApp.toolbox.OSToolbox;
     case 'com.endlessm.Sidetrack':
         return imports.RileyMaze.toolbox.RMZToolbox;
+    case 'name.ptomato.AbsentRhythm':
+        return imports.sketchbook.toolbox.SketchToolbox;
     default:
         return imports.hacktoolbox.hacktoolbox.DefaultHackToolbox;
     }
--- a/src/codeview.js
+++ b/src/codeview.js
@@ -132,6 +132,10 @@ const Buffer = GObject.registerClass(class Buffer extends GtkSource.Buffer {
 
 var Codeview = GObject.registerClass({
     GTypeName: 'Codeview',
+    Properties: {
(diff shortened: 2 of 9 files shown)


src/codeview.js
```diff
@@ -232,6 +239,7 @@ var Codeview = GObject.registerClass({
             });
             this._cached_ast = null;
             this.compile(false);
+            this.notify('text');
```
manuq, line 242: The notify is automatic at least in python, unless you add the`GObject.ParamFlags.EXPLICIT_NOTIFY` flag to the property. Isn't the same in GJS?

src/codeview.js
```diff
@@ -223,6 +227,9 @@ var Codeview = GObject.registerClass({
     }
 
     set text(value) {
+        if (this._buffer && this._buffer.text === value)
+            return;
```
manuq, line 231: Maybe you need the `GObject.ParamFlags.EXPLICIT_NOTIFY` flag for this to return without notifying?

src/app.js
```diff
@@ -30,6 +30,8 @@ function _toolboxClassForAppId(targetAppId) {
         return imports.OperatingSystemApp.toolbox.OSToolbox;
     case 'com.endlessm.Sidetrack':
         return imports.RileyMaze.toolbox.RMZToolbox;
+    case 'name.ptomato.AbsentRhythm':
```
manuq: Now this could be the actual name, 'com.endlessm.Sketchbook'.

src/sketchbook/model.js
```diff
@@ -1,29 +1,142 @@
 /* exported SketchModel */
 
-const {GObject} = imports.gi;
+const {Gio, GObject} = imports.gi;
+
+const INTERFACE = `
+<node>
+  <interface name="name.ptomato.AbsentRhythm.Variables">
```
manuq: This need to be 'com.endlessm.Sketchbook' now.

src/sketchbook/model.js
```diff
@@ -1,29 +1,142 @@
...
-            ''),
+            GObject.ParamFlags.READWRITE, ''),
     },
 }, class SketchModel extends GObject.Object {
+    // This ought to be implemented via the Gio.AsyncInitable interface, but
+    // that's not currently possible in GJS
+    async initAsync(cancellable = null) {
+        const sessionBus = await Gio.DBus.get(Gio.BusType.SESSION, cancellable);
+        this._proxy = await new Promise((resolve, reject) => {
+            void new VariablesProxy(sessionBus, 'name.ptomato.AbsentRhythm',
```
manuq: Ditto, rename needed.

manuq (changes requested):
Looking good but I couldn't test it yet. Leaving a few comments to ;keep the rock rolling.

src/codeview.js
ptomato: It's automatic if you use [`GObject.set_property()`](https://gjs-docs.gnome.org/gobject20~2.0_api/gobject.object#method-set_property), but manual if you use JS property setters.

manuq: oh. Ok.

manuq (approved):
Thanks for clarifying and the rename. I haven't tried it but I'll merge as-is since you are moving to the Platform team, and there are no plans to this prototype right now, but yes in the future.

PR #227, opened 2019-07-03
window: Add topbar for HackUnlock toolbox window

cfoch:
https://phabricator.endlessm.com/T27142

### DIFF
--- a/src/window.js
+++ b/src/window.js
@@ -3,6 +3,29 @@
 const {Gio, GLib, GObject, Gtk, HackToolbox} = imports.gi;
 const {Lockscreen} = imports.lockscreen;
 
+const DATA_RESOURCE_PATH = 'resource:///com/endlessm/HackToolbox';
+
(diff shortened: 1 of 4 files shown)


src/window.js
```diff
@@ -46,7 +69,17 @@ var ToolboxWindow = GObject.registerClass({
         });
         this._toolbox_frame.get_style_context().add_class('toolbox');
 
-        Gtk.Container.prototype.add.call(this, this._lockscreen);
+        let container = this._lockscreen;
+        if (GLib.getenv('HACK_TOOLBOX_HACK_MODE_ENABLED') &&
+                this.target_app_id == 'com.endlessm.HackUnlock') {
```
danigm: Let's fix the eslint problems before merge, you should use `===` instead of `==`.

danigm (changes requested):

danigm (approved):

PR #233, opened 2019-09-12
OperatingSystemApp: rotate cursor glitchy

danigm:
The cursor upside-down is not very useful, it's a bit confuse to use it.
In the clubhouse we're using the cursor-glitchy, but rotated to look
like a normal cursor and the name colides with this.

This patch uses the same cursor in the OperatingSystemApp toolbox so the
original is not overriden.

https://phabricator.endlessm.com/T27517


cfoch (approved):

danigm:
Closing old PR

PR #236, opened 2019-12-10
Sketchbook: Use the new com.hack_computer prefix

danigm:
https://phabricator.endlessm.com/T28735


manuq (approved):
The replacements look good!

PR #239, opened 2020-01-10
playbin: Gst playbin properties in C

danigm:
There's a problem with flatpak platforms bigger than 2.28 and accessing
gst playbin properties. This patch is a workaround to get/set properties
of the lock screen playbin widget from javascript, just creating the
methods in C.

https://phabricator.endlessm.com/T29102

### DIFF
--- a/src/lib/htb-playbin.c
+++ b/src/lib/htb-playbin.c
@@ -1,0 +1,51 @@
+#include "htb.h"
+
+/**
+ * hack_toolbox_playbin_get_widget:
+ * @obj: The parent object
+ *
+ * Get the widget property of a gstbin object. This is a workaround to get
+ * the playbin widget in javascript because the direct access in gjs is not
+ * working since org.gnome.Platform//2.28
+ *
+ * Returns: (transfer full): a #GtkWidget
+ */
+GtkWidget *
+hack_toolbox_playbin_get_widget(GObject *obj)
+{
+  GtkWidget *widget = NULL;
+  GObject * sink = NULL;
+  g_object_get (obj, "video_sink", &sink, NULL);
+  g_object_get (sink, "widget", &widget, NULL);
+  return widget;
+}
+
+/**
+ * hack_toolbox_playbin_set_uri:
+ * @obj: The gstbin
+ *
+ * Set the uri property of a gstbin object. This is a workaround to set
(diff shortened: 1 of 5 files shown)


src/lib/htb-playbin.c
```diff
@@ -0,0 +1,51 @@
...
+ * @obj: The parent object
+ *
+ * Get the widget property of a gstbin object. This is a workaround to get
+ * the playbin widget in javascript because the direct access in gjs is not
+ * working since org.gnome.Platform//2.28
+ *
+ * Returns: (transfer full): a #GtkWidget
+ */
+GtkWidget *
+hack_toolbox_playbin_get_widget(GObject *obj)
```
xjuan, line 14: I would make this more generic, just a wrapper for gobject properties
`GObject* hack_toolbox_get_object(obj, property)`

src/lib/htb-playbin.c
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * hack_toolbox_playbin_set_uri:
+ * @obj: The gstbin
+ *
+ * Set the uri property of a gstbin object. This is a workaround to set
+ * the playbin uri in javascript because the direct access in gjs is not
+ * working since org.gnome.Platform//2.28
+ */
+void
+hack_toolbox_playbin_set_uri(GObject *obj,
```
xjuan, line 32: set_property() seems to work so this would not be needed

src/lib/htb-playbin.c
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * hack_toolbox_playbin_set_video_filter:
+ * @obj: The gstbin
+ *
+ * Set the uri property of a gstbin object. This is a workaround to set
+ * the playbin uri in javascript because the direct access in gjs is not
+ * working since org.gnome.Platform//2.28
+ */
+void
+hack_toolbox_playbin_set_video_filter(GObject *obj,
```
xjuan, line 47: Same here, I think set_property('video-filter', filter) should work

xjuan (commented):
I think this can be simplified quite a bit by just introducing one function to get object properties.

See https://github.com/endlessm/hack-toolbox-app/pull/240

danigm:
Closing old PR

PR #240, opened 2020-01-10
playbin: workaround GJS regression

xjuan:
There is a regression that is preventing setting/getting properties
from GStreamer objects using JS accessors.

We workaround this by use gobject set_property() functions and
a new custom symbols for getting object properties

https://phabricator.endlessm.com/T29102


danigm (approved):
LGTM, it's a simple workaround and solves the problem meanwhile we wait for the fix in gjs.

PR #241, opened 2020-01-24
codeview: Add codeview dbus interface to expose code errors

danigm:
https://phabricator.endlessm.com/T29202

### DIFF
--- a/src/codeview.js
+++ b/src/codeview.js
@@ -6,6 +6,13 @@ const ByteArray = imports.byteArray;
 const SoundServer = imports.soundServer;
 void imports.utils;  // pull in promisified functions
 
+const CodeViewIface = `
+<node>
+  <interface name="com.hack_computer.HackToolbox.CodeView">
+    <property name="errors" type="b" access="read"/>
+  </interface>
+</node>`;
+
 const CODEVIEW_ERROR_EVENT = 'e98aa2b8-3f11-4a25-b8e9-b10a635df121';
 
 // Can add more, e.g. WARNING, SUGGESTION
@@ -135,6 +142,9 @@ var Codeview = GObject.registerClass({
     Properties: {
         text: GObject.ParamSpec.string('text', 'Text', 'Source code being edited',
             GObject.ParamFlags.READWRITE, ''),
+        errors: GObject.ParamSpec.boolean('errors', 'Errors',
+            'True if there are errors',
+            GObject.ParamFlags.READABLE, false),
     },
     Signals: {
         'should-compile': {
@@ -154,6 +164,7 @@ var Codeview = GObject.registerClass({
         const schemeManager = GtkSource.StyleSchemeManager.get_default();
         const styleScheme = schemeManager.get_scheme('hack');
 
+        this._dbus = null;
         this._buffer = new Buffer({language, styleScheme});
 
         const bgErr = new Gdk.RGBA();
@@ -218,6 +229,10 @@ var Codeview = GObject.registerClass({
         // This is the identifier of the user function being edited in this
         // codeview. Empty string is the regular variable assignment panel.
         this._userFunction = '';
+    }
+
+    get errors() {
+        return !!this._numErrors;
     }
 
     get text() {
@@ -556,6 +571,24 @@ var Codeview = GObject.registerClass({
         }
     }
 
+    dbusRegister(appId, id) {
+        const objPath = Gio.Application.get_default().get_dbus_object_path();
+        const winName = appId.replace(/\./gi, '_');
+        this.objectPath = `${objPath}/window/${winName}/codeview/${id}`;
+        this._dbus = Gio.DBusExportedObject.wrapJSObject(CodeViewIface, this);
+        try {
+            this._dbus.export(Gio.DBus.session, this.objectPath);
+        } catch (e) {
+            this._dbus = null;
+            logError(e);
+        }
+    }
+
+    dbusUnregister() {
+        if (this._dbus)
+            this._dbus.unexport();
+    }
+
     setCompileResults(results, userInitiated = true) {
         const [begin, bound] = this._buffer.get_bounds();
         this._buffer.remove_source_marks(begin, bound, MarkType.ERROR);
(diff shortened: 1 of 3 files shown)


src/codeview.js
```diff
@@ -135,6 +142,9 @@ var Codeview = GObject.registerClass({
     Properties: {
         text: GObject.ParamSpec.string('text', 'Text', 'Source code being edited',
             GObject.ParamFlags.READWRITE, ''),
+        errors: GObject.ParamSpec.uint('errors', 'Errors',
```
cfoch: There is already a `this._numErrors` attribute. You should use that or rename it.

src/codeview.js
```diff
@@ -556,10 +572,35 @@ var Codeview = GObject.registerClass({
         }
     }
 
+    dbusRegister(appId, id) {
+        const objPath = Gio.Application.get_default().get_dbus_object_path();
+        const winName = appId.replace(/\./gi, '_');
+        this.objectPath = `${objPath}/window/${winName}/codeview/${id}`;
+        this.address = `com.hack_computer.HackToolbox.window.${winName}.codeview.${id}`;
```
cfoch: I think this is not used?

src/codeview.js
```diff
@@ -556,10 +572,35 @@ var Codeview = GObject.registerClass({
         }
     }
 
+    dbusRegister(appId, id) {
+        const objPath = Gio.Application.get_default().get_dbus_object_path();
+        const winName = appId.replace(/\./gi, '_');
```
cfoch, line 576: I was thinking that we should use the window id here just to follow the same object path pattern used in the ToolboxWindow, but I am not sure how you could know the toolbox window id from the Clubhouse. If you find the way, better to do that, IMO.

src/codeview.js
```diff
@@ -556,10 +572,35 @@ var Codeview = GObject.registerClass({
...
+            this._dbus.unexport();
+    }
+
     setCompileResults(results, userInitiated = true) {
         const [begin, bound] = this._buffer.get_bounds();
         this._buffer.remove_source_marks(begin, bound, MarkType.ERROR);
         this._buffer.remove_tag(this._errorStyle, begin, bound);
+
+        if (this._dbus) {
+            this._errors = results.length;
```
cfoch: If we have `errors` as a GObject property, we can just call `emit_property_changed` on the setter or connecting to `notify::errors` and do it on its handler.

src/codeview.js
```diff
@@ -154,6 +164,8 @@ var Codeview = GObject.registerClass({
         const schemeManager = GtkSource.StyleSchemeManager.get_default();
         const styleScheme = schemeManager.get_scheme('hack');
 
+        this._dbus = null;
+        this._errors = 0;
```
cfoch: There is already a `this._numErrors = 0`. It will be confusing to have both.

src/codeview.js
```diff
@@ -135,6 +142,9 @@ var Codeview = GObject.registerClass({
     Properties: {
         text: GObject.ParamSpec.string('text', 'Text', 'Source code being edited',
             GObject.ParamFlags.READWRITE, ''),
+        errors: GObject.ParamSpec.uint('errors', 'Errors',
+            'Number of errors in this codeview',
+            GObject.ParamFlags.READWRITE, 0, 500, 0),
```
cfoch: `GLib.MAXUINT32` instead of 500?

cfoch (changes requested):

danigm:
Thanks for the review, I've applied all suggested changes.

src/codeview.js
```diff
@@ -585,7 +626,8 @@ var Codeview = GObject.registerClass({
         });
 
         if (this._numErrors !== results.length) {
```
cfoch, line 620: Instead of accessing `this._numErrors` directly, use `this.errors`. Here and everywhere `this._numErrors` appears.

src/codeview.js
```diff
@@ -556,10 +579,28 @@ var Codeview = GObject.registerClass({
...
+    dbusUnregister() {
+        if (this._dbus)
+            this._dbus.unexport();
+    }
+
     setCompileResults(results, userInitiated = true) {
         const [begin, bound] = this._buffer.get_bounds();
         this._buffer.remove_source_marks(begin, bound, MarkType.ERROR);
         this._buffer.remove_tag(this._errorStyle, begin, bound);
+
```
cfoch: remove added line

src/codeview.js
```diff
@@ -556,10 +579,28 @@ var Codeview = GObject.registerClass({
...
 
+    dbusRegister(appId, id) {
+        const objPath = Gio.Application.get_default().get_dbus_object_path();
+        const winName = appId.replace(/\./gi, '_');
+        this.objectPath = `${objPath}/window/${winName}/codeview/${id}`;
+        this._dbus = Gio.DBusExportedObject.wrapJSObject(CodeViewIface, this);
+        try {
+            this._dbus.export(Gio.DBus.session, this.objectPath);
+        } catch (e) {
+            logError(e);
```
cfoch, line 583: If we cannot export, wouldn't it make sense to set `this._dbus` to null here?

cfoch (changes requested):

src/codeview.js
```diff
@@ -220,6 +231,18 @@ var Codeview = GObject.registerClass({
         this._userFunction = '';
     }
 
+    get errors() {
+        return this._numErrors;
+    }
+
+    set errors(value) {
+        this._numErrors = value;
```
cfoch: Return/skip if `this._numErrors === value`, otherwise a property change will be emitted in vain.

cfoch (changes requested):

cfoch (commented):
You also have lint errors:

```
/src/src/codeview.js
  665:1  error  Line 665 exceeds the maximum comment line length of 80  max-len
```

src/codeview.js
danigm: ok

src/codeview.js
danigm: yes, sure, I've just added that

src/codeview.js
```diff
@@ -586,6 +619,12 @@ var Codeview = GObject.registerClass({
 
         if (this._numErrors !== results.length) {
             this._numErrors = results.length;
+
+            if (this._dbus) {
+                const variant = new GLib.Variant('b', !!this._numErrors);
+                this._dbus.emit_property_changed('errors', variant);
```
cfoch, line 625: Notify the GObject property change. too.

cfoch: `this.notify('errors');`'

cfoch (changes requested):

src/codeview.js
danigm: done

cfoch (approved):

PR #244, opened 2020-03-09
OperatingSystemApp: Use extension dbus instead shell settings

danigm:
https://phabricator.endlessm.com/T28568

### Notes

This depends on the hack code move to a shell extension so don't merge before this one: https://github.com/endlessm/gnome-shell/pull/631


manuq (approved):
Looks good as well! According to the CI check, eslint is failing. But has nothing to do with these changes:
```
/src/src/Fizzics/level2.js
  98:27  error  Expected 'this' to be used by class method '_getValueForScopeValue'  class-methods-use-this
```

PR #247, opened 2020-08-10
Use new names for sidetrack and OS

danigm:
https://phabricator.endlessm.com/T30458


danigm:
To bundle the toy apps inside the clubhouse katamari is not as easy as expected. The gnome-shell alt-tab, panel and dash manages all app windows inside the same flatpak as part of the same app, so it groups the toy app window with the clubhouse window and we don't want that behavior.

PR #250, opened 2020-11-17
Use clubhouse recordMetric instead of eosmetics

danigm:
This PR requires the API in the clubhouse: https://github.com/endlessm/clubhouse/pull/1198

We are replacing the metrics system on the clubhouse to use a generic
matomo server instead of the eosmetrics. This patch removes the
dependency from the toolbox and uses the new clubhouse API to register
codeview error metrics.

https://phabricator.endlessm.com/T30942


manuq (commented):
Looks good! Please merge after merging the clubhouse PR.

PR #254, opened 2022-03-18
Use hack properties for cursor config

danigm:
It's not possible anymore to modify general settings from the flatpak
sandbox, so instead of doing that directly to update the cursor-theme,
we can use the new interface provided by the shell hack extension.

https://phabricator.endlessm.com/T33281

This change depends on this PR: https://github.com/endlessm/eos-hack-extension/pull/57

### DIFF
--- a/src/OperatingSystemApp/oscursormodel.js
+++ b/src/OperatingSystemApp/oscursormodel.js
@@ -60,47 +60,11 @@ var OSCursorModel = GObject.registerClass({
 
         super._init();
 
-        var ifaceSettings = new Gio.Settings({
-            schemaId: 'org.gnome.desktop.interface',
-        });
-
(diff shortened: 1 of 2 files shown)


src/OperatingSystemApp/oscursormodel.js
```diff
@@ -110,8 +74,8 @@ var OSCursorModel = GObject.registerClass({
          * up to 256px
          */
         this.theme = 'cursor-default';
-
-        this._touchpadSettings.reset('speed');
+        this.size = 24;
+        this.speed = 0;
```
dylanmccall, line 78: It would be nice if this was using a constant defined somewhere, but we have a big list of magic numbers with all the `GObject.ParamSpec` calls here as well so :man_shrugging:

dylanmccall (approved):
:+1:

